PR #62, opened 2019-07-16
migrator: migrate parent records init, migrate documents

kprzerwa:
* add migration module

* migrates parent records (serials) from json files

* migrates documents "on the fly" without creating files
 (needs xml file source)

* migration is not linking records

(to discuss) I will describe the migration process in the internal docs


cds_books/migrator/cli.py
```diff
@@ -0,0 +1,115 @@
...
+                            "migration: report number associated with multiple"
+                            "recid. See {0}".format(item['recid']))
+
+
+@dumps.command()
+@click.argument('sources', type=click.File('r'), nargs=-1)
+@click.option(
+    '--source-type',
+    '-t',
+    type=click.Choice(['json', 'marcxml']),
```
ntarocco: Mark?

cds_books/migrator/records.py
```diff
@@ -0,0 +1,196 @@
...
+        record_uuid = uuid.uuid4()
+        provider = model._pid_provider.create(
+            object_type='rec',
+            object_uuid=record_uuid,
+        )
+        dump[model.pid_field] = provider.pid.pid_value
+        record = model.create(dump, record_uuid)
+        record.model.created = datetime.datetime.now()
+        record.commit()
+        return record
```
ntarocco: Is there any indexing happening at some point?

cds_books/migrator/records.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class CDSRecordDumpLoader(RecordDumpLoader):
+    """Migrate a CDS record."""
+
+    dependent_objs = ['frame', 'frames-preview', 'playlist', 'subformat']
```
ntarocco: Videos leftovers? :)

cds_books/migrator/cli.py
```diff
@@ -0,0 +1,115 @@
...
+from invenio_records import Record
+
+from cds_books.migrator.records import CDSParentRecordDumpLoader
+
+
+def load_dumps_from_dir(source_dir, rectype):
+    with click.progressbar(os.listdir(source_dir)) as bar:
+        for filename in bar:
+            click.echo('Loading dump {0}'.format(filename))
+            if re.match(r'^'+rectype+'_(\d+)(_\d+)*.json$', filename):
```
equadon: Now we changed the migrator to store all dumps in a single file so this will break. Should we change it back to store each dump in its own file or keep it as-is in a single file?

cds_books/migrator/cli.py
```diff
@@ -0,0 +1,115 @@
...
+@click.option(
+    '--recid',
+    '-r',
+    help='Record ID to load (NOTE: will load only one record!).',
+    default=None)
+@click.option(
+    '--rectype',
+    '-x',
+    help='Type of record to load (f.e serial)',
+    default=None)
```
equadon: The default in migrator kit has been changed to `document` instead of `None` to be more explicit but not sure if that will have an effect here.

cds_books/migrator/cli.py
equadon: The `import_record` task we're using only supports `marcxml` and `json`.

cds_books/migrator/records.py
equadon: Indexing has now been implemented for migrating parents, documents and relations.

Pipfile
```diff
@@ -11,7 +11,15 @@ uwsgitop = ">=0.11"
 uwsgi-tools = ">=1.1.1"
 lxml = ">=3.5.0,<4.2.6"
 invenio-app-ils = {editable = true,git = "https://github.com/inveniosoftware/invenio-app-ils.git",ref = "master",extras = [ "lorem" ]}
+invenio-migrator = {editable = true,version = "==1.0.0a10"}
+cds-books = {editable = true,path = "."}
+invenio-db = {editable = true,version = "*"}
+invenio-files-rest = "*"
+invenio-records-files = "*"
+cds-dojson = {editable = true,git = "https://github.com/CERNDocumentServer/cds-dojson",ref = "books"}
```
ntarocco: I would really prefer to depend on a release, we might forget.
Or we really have to remember to fix it afterwards!

PR #75, opened 2019-09-11
global: fix travis errors

equadon:
* Fixes `numpy` error on Travis
* Remove E2E tests


run-tests.sh
```diff
@@ -9,7 +9,7 @@
 # Ignoring false positive 36759 (reporting invenio-admin v1.0.1). This can be
 # removed when https://github.com/pyupio/safety-db/pull/2274 is merged and
 # released.
-pipenv check --ignore 36759 && \
+pipenv check --ignore 36759 --ignore 36810 && \
```
zzacharo: @equadon can you put just a small note about `36810` so we can have it in mind in the future to check the issue if it is fixed?

PR #76, opened 2019-09-11
patrons: retrieve patron loans information

zzacharo:


cds_books/patrons/api.py
```diff
@@ -0,0 +1,59 @@
...
+        """Return python representation of Patron metadata."""
+        dump = {
+            "$schema": self._schema,
+            "id": self.id,
+            "name": self.profile.full_name if self.profile else "",
+            "email": self.user.email,
+        }
+        if self.extra_info:
+            dump.update({
+                "person_id": self.extra_info["person_id"],
```
ntarocco: Maybe safer to do:
```
  "person_id": self.extra_info.get("person_id", ""),
  "department": self.extra_info.get("department", "")
```

cds_books/patrons/views.py
```diff
@@ -0,0 +1,88 @@
...
+
+from .serializers import patron_loans_serializer
+
+
+def pass_patron_from_es():
+    """Decorator to retrieve a patron from ES."""
+    def pass_patron_decorator(f):
+        @wraps(f)
+        def inner(self, person_id, *args, **kwargs):
+            results = PatronsSearch().filter(
```
ntarocco: Does Flask with `string:patron_id` validates already if you put some weird thing? Sanitize the param?

cds_books/patrons/views.py
```diff
@@ -0,0 +1,88 @@
...
+    @need_permissions("retrieve-patron-loans")
+    @pass_patron_from_es()
+    def get(self, patron, **kwargs):
+        """Retrieve patron loans."""
+        active_loan_states = current_app.config[
+            "CIRCULATION_STATES_LOAN_ACTIVE"]
+        active_loans = search_by_patron_item_or_document(
+            patron["id"], filter_states=active_loan_states).execute()
+
+        pending_loan_states = ["CREATED", "PENDING"]
```
ntarocco: Hopefully we have nothing in `CREATED`. Actually, maybe better to return only `PENDING` because if for any reason we have some weird loans in CREATED state that are wrong, we expose them, WDYT?

cds_books/config.py
```diff
@@ -166,8 +172,9 @@ def _parse_env_bool(var_name, default=None):
 #: should be set to the correct host and it is strongly recommended to only
 #: route correct hosts to the application.
 APP_ALLOWED_HOSTS = [
-    os.environ.get("ALLOWED_HOST"),
-    os.environ.get("HOSTNAME"),  # fix disallowed host error during /ping
+    # os.environ.get("ALLOWED_HOST"),
+    # os.environ.get("HOSTNAME"),  # fix disallowed host error during /ping
+    "localhost"
```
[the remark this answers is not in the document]
zzacharo: oups :)

cds_books/patrons/api.py
```diff
@@ -0,0 +1,59 @@
...
+
+from __future__ import absolute_import, print_function
+
+from flask import current_app
+from invenio_accounts.models import User
+from invenio_oauthclient.models import RemoteAccount
+from invenio_userprofiles.api import UserProfile
+
+
+class Patron(object):
```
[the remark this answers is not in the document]
zzacharo: I thought about that, ok I will do it!

cds_books/patrons/views.py
zzacharo: I will check that!

cds_books/patrons/views.py
zzacharo: Makes sense, I just copied what it was used already in ils ;)

cds_books/patrons/views.py
zzacharo: I made the converter to accept only integers for person_id. Flask says that if is not a valid integer it results to a 404! http://exploreflask.com/en/latest/views.html#built-in-converters

PR #79, opened 2019-09-24
cli: index ldap users after import

zzacharo:


cds_books/ldap/api.py
```diff
@@ -175,8 +177,10 @@ def _commit_user(user_data):
             return user.id
 
         for ldap_user in self.ldap_users:
-            user_id = _commit_user(ldap_user)
+            print("Importing user with person id {}".format(
```
ntarocco: I guess `print` was really wanted right?

zzacharo: I didn't have any logger setup so it was the fastest option for now. There is an issue for adding proper logging :)

PR #85, opened 2019-10-14
global: update ils to spa deployment

zzacharo:


Pipfile
```diff
@@ -10,7 +10,7 @@ uwsgi = ">=2.0"
 uwsgitop = ">=0.11"
 uwsgi-tools = ">=1.1.1"
 lxml = ">=3.5.0,<4.2.6"
-invenio-app-ils = {editable = true,git = "https://github.com/inveniosoftware/invenio-app-ils.git",ref = "master",extras = [ "lorem" ]}
+invenio-app-ils = {editable = true,git = "https://github.com/inveniosoftware/invenio-app-ils.git",ref = "master", extras = [ "lorem" , "elasticsearch6" ]}
```
ntarocco: since with this [PR](https://github.com/inveniosoftware/invenio-app-ils/pull/546/files#diff-2eeaed663bd0d25b7e608891384b7298R43) there will be also the db selection, can you please add `postgresql` too so we don't need another change later on?
It should not fail I guess!

zzacharo: ok!

Issue #88, opened 2020-01-06
Openshift: stats not calculated properly


kprzerwa:
- [x] fix statistics on books-dev ~(#89)~ (inveniosoftware/invenio-stats#111)
- [ ] enable automatic deployment when tags pushed

Requires #87 (see comment below)

equadon:
Some work have been made to address the stats issue on `books-dev` in https://github.com/CERNDocumentServer/cds-books/pull/89 and in `cds-books-openshift`. However, due to missing #87 we can't test statistics properly since `invenio-stats` returns `null` if not both record view and download stats are working properly.

PR #89, opened 2020-01-07
global: fix build issues

equadon:
* Updated to latest ILS
* ~~Use webargs < 6.0~~
* Fix build and deployment errors
* Possible fix for stats (#88) (not yet verified)

The operations documentation for books have also been updated with the new build/deployment configuration added for the OpenShift overlay.


cds_books/config.py
```diff
@@ -142,6 +142,16 @@ def _parse_env_bool(var_name, default=None):
         'task': 'invenio_accounts.tasks.clean_session_table',
         'schedule': timedelta(minutes=60),
     },
+    "stats-process-events": {
```
ntarocco: I think we should change this: instead of re-defining CELERY_BEAT_SCHEDULE, we should extend it with the cds books config.

cds_books/config.py
```diff
@@ -135,16 +136,10 @@ def _parse_env_bool(var_name, default=None):
...
 CELERY_BEAT_SCHEDULE = {
-    'indexer': {
-        'task': 'invenio_indexer.tasks.process_bulk_queue',
-        'schedule': timedelta(minutes=5),
-    },
-    'accounts': {
-        'task': 'invenio_accounts.tasks.clean_session_table',
-        'schedule': timedelta(minutes=60),
-    },
+    **ILS_CELERY_BEAT_SCHEDULE,
```
ntarocco, line 139: You can remove this if there is no change: `CELERY_BEAT_SCHEDULE` will be taken from ILS.

PR #93, opened 2020-04-23
syndetic covers

### LINKED ISSUE #92: Syndetic book cover integration
Integrate Syndetic specifics with ILS
- [x] When a Document is created/updated, the `cover_metadata` should be updated with the potentially changed ISBN. This could be done in a cron job (task below) but it will add a new commit to the record, a new version. We could try to guess it first so that the cron job might do nothing:
  - Connect to the before [create signal](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L175) and before [commit signal](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L263) only for Document record, no other record types
  - In the implementation, if `cover_metadata` already contains an ISBN, do nothing. If does not contain an ISBN, put the first (hopefully is the working one!)
- [x]  Set the working ISBN in `cover_metadata`. In CDS-ILS ISBN are stored only on Documents, no need to check for Items. Create an `async task` that will run after some time after the creation/edition of a Document.
  - Connect to the after [create signal](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L186) and after [commit signal](https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L275) only for Document record, no other record types
  - if `cover_metadata` contains already one ISBN, test it and if it succeeds just exit. If `cover_metadata` does not contain am `ISBN` or it the existing one does not work, get all ISBNs from the `identifiers` field and try to find the one that works, then set it in `cover_metadata`.
- [x] Implement `cover_builder` for `ILS_BUILD_LITERATURE_COVER_URLS`  override in ILS.

### Syndetics' URL Description

1.  the item’s ISBN
1. the filename of the data element being requested
1. the library’s identification code

## Resources

- [Syndetic Provider Documentation](http://corpweb.proquest.com/docs/syndeticsstarterdocument.pdf)
- Existing [integration](https://gitlab.cern.ch/AIGROUP-cds-admin/cds-legacy/-/commit/baa17c0fbe40c5b7da8e6b516da1c727a3218e55#2775be67a5d59942677f83a6deee6edaa1e43bb8_525_531) of Syndetic book cover provider with CDS.

related inveniosoftware/invenio-app-ils#560

topless:
- Syndetic covers integration 
- [documentation](https://github.com/inveniosoftware/docs-invenio-ils/pull/1)
- [openshift configuration](https://gitlab.cern.ch/invenio/template-openshift/-/merge_requests/8)
- related to https://github.com/inveniosoftware/invenio-app-ils/pull/841
- closes #92

### DIFF
--- a/cds_books/config.py
+++ b/cds_books/config.py
@@ -21,6 +21,7 @@ from datetime import timedelta
 from invenio_app_ils.config import RECORDS_REST_ENDPOINTS
 from invenio_app_ils.pidstore.pids import PATRON_PID_TYPE
 
+from .literature.covers import build_literature_cover_urls
 from .patrons.api import Patron
 from .patrons.permissions import views_permissions_factory
 
@@ -240,3 +241,5 @@ MIGRATOR_RECORDS_DUMP_CLS = 'cds_books.migrator.records:CDSRecordDump'
 JSONSCHEMAS_SCHEMAS = ['ils', 'loans', 'documents', "document_requests",
                        "acquisition", "ill", "invenio_records_files",
                        "invenio_opendefinition"]
+
+ILS_LITERATURE_COVER_URLS_BUILDER = build_literature_cover_urls
--- a/cds_books/config.py
+++ b/cds_books/config.py
@@ -21,7 +21,7 @@ from datetime import timedelta
 from invenio_app_ils.config import RECORDS_REST_ENDPOINTS
 from invenio_app_ils.pidstore.pids import PATRON_PID_TYPE
 
-from .literature.covers import build_literature_cover_urls
+from .literature.covers import build_syndetic_cover_urls
 from .patrons.api import Patron
 from .patrons.permissions import views_permissions_factory
 
@@ -238,8 +238,8 @@ MIGRATOR_RECORDS_DUMPLOADER_CLS = \
     'cds_books.migrator.records:CDSDocumentDumpLoader'
 MIGRATOR_RECORDS_DUMP_CLS = 'cds_books.migrator.records:CDSRecordDump'
 
-JSONSCHEMAS_SCHEMAS = ['ils', 'loans', 'documents', "document_requests",
+JSONSCHEMAS_SCHEMAS = ["ils", "loans", "documents", "document_requests",
                        "acquisition", "ill", "invenio_records_files",
                        "invenio_opendefinition"]
 
-ILS_LITERATURE_COVER_URLS_BUILDER = build_literature_cover_urls
+ILS_LITERATURE_COVER_URLS_BUILDER = build_syndetic_cover_urls
--- a/cds_books/literature/__init__.py
+++ b/cds_books/literature/__init__.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019-2020 CERN.
+#
+# CDS Books is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""CDS Books literature module."""
--- a/cds_books/literature/covers.py
+++ b/cds_books/literature/covers.py
@@ -8,6 +8,7 @@
 """Literature covers."""
 
 import os
+from functools import partial
 
 from invenio_app_ils.literature.covers_builder import build_placeholder_urls
 
@@ -15,9 +16,8 @@ from invenio_app_ils.literature.covers_builder import build_placeholder_urls
 def is_record_with_cover(record):
     """Check if this type of record has cover."""
     schema = record.get("$schema")
-    if (
-        schema.endswith("document-v1.0.0.json") or
-        schema.endswith("series-v1.0.0.json")
+    if schema.endswith("document-v1.0.0.json") or schema.endswith(
+        "series-v1.0.0.json"
     ):
         return True
     return False
@@ -26,8 +26,8 @@ def is_record_with_cover(record):
 def has_already_cover(record):
     """Check if record has already valid cover in cover_metadata."""
     cover_metadata = record.get("cover_metadata", {})
-    return (
-        cover_metadata.get("ISBN", False) or cover_metadata.get("ISSN", False)
+    return cover_metadata.get("ISBN", False) or cover_metadata.get(
+        "ISSN", False
     )
 
 
@@ -68,21 +68,27 @@ def build_syndetic_cover_urls(metadata):
 
     issn = cover_metadata.get("ISSN", "")
     if issn:
-        scheme = 'ISSN'
+        scheme = "ISSN"
         scheme_value = issn
 
     isbn = cover_metadata.get("ISBN", "")
     if isbn:
-        scheme = 'ISBN'
+        scheme = "ISBN"
         scheme_value = isbn
 
     if issn or isbn:
+        _url = "{url}?client={client}&{sheme}={value}/{size}.gif"
+        partial_url = partial(
+            _url.format,
+            url=url,
+            client=client,
+            sheme=scheme,
+            value=scheme_value,
+        )
         return {
-            "small": "{url}?client={client}&{sheme}={scheme_value}/SC.gif".format(
-                url=url, client=client, sheme=scheme, scheme_value=scheme_value),
-            "medium": "{url}?client={client}&{sheme}={scheme_value}/MC.gif".format(
-                url=url, client=client, sheme=scheme, scheme_value=scheme_value),
-            "large": "{url}?client={client}&{sheme}={scheme_value}/LC.gif".format(
-                url=url, client=client, sheme=scheme, scheme_value=scheme_value),
+            "is_placeholder": False,
+            "small": partial_url(size="SC"),
+            "medium": partial_url(size="MC"),
+            "large": partial_url(size="LC"),
         }
     return build_placeholder_urls()
--- a/cds_books/literature/covers.py
+++ b/cds_books/literature/covers.py
@@ -7,51 +7,18 @@
 
 """Literature covers."""
 
-from invenio_app_ils.documents.covers_builder import build_default_cover_urls
 
-
-def build_literature_cover_urls(record):
+def build_literature_cover_urls(metadata):
     """Decorate literature with cover urls for all sizes."""
-
-    def build_urls(url, params):
-        """Build all available size urls."""
-        return {
-            "small": "{0}/{1}".format(url, params["small"]),
-            "medium": "{0}/{1}".format(url, params["medium"]),
-            "large": "{0}/{1}".format(url, params["large"]),
-        }
-
-    syndetic_client = "cernlibrary"
-    syndetic_domain = "https://secure.syndetics.com/index.aspx"
-    syndentic_sizes = {
-        "small": "SC.gif",
-        "medium": "MC.gif",
-        "large": "MC.gif",
+    client = "cernlibrary"
+    url = "https://secure.syndetics.com/index.aspx"
+    cover_metadata = metadata.get("cover_metadata", {})
+    isbn = cover_metadata.get("isbn", "")
+    return {
+        "small": "{url}?client={client}&isbn={isbn}/SC.gif".format(
+            url=url, client=client, isbn=isbn),
+        "medium": "{url}?client={client}&isbn={isbn}/MC.gif".format(
+            url=url, client=client, isbn=isbn),
+        "large": "{url}?client={client}&isbn={isbn}/LC.gif".format(
+            url=url, client=client, isbn=isbn),
     }
-
-    result = build_default_cover_urls(record)
-
-    if (
-        "items" in record and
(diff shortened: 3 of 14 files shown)


cds_books/literature/covers.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# CDS Books is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Literature covers."""
+
+
+def build_literature_cover_urls(metadata):
+    """Decorate literature with cover urls for all sizes."""
+    client = "cernlibrary"
```
ntarocco, line 13: I would say that we should have this is a config/secret... it is not a secret because it is in the URL, but it is kind of bothering me! WDYT?

cds_books/literature/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: minor: can we name the package plural? I saw now that it is not in ILS, but I/we will change it since we named all records' modules plural.

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+ALLOWED_CLASSES = [Document]
+MIN_CONTENT_LENGTH = 128
+
+
+def update_cover(sender, *args, **kwargs):
+    """Triggers async task to update cover metadata of the record."""
+    record = kwargs.get("record", {})
+    if record.__class__ in ALLOWED_CLASSES:
```
ntarocco, line 32: if you have the schema field, it is better to check that, because you might override the class.

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+
+from celery import shared_task
+from invenio_app_ils.documents.api import Document
+from invenio_db import db
+
+from .covers import build_literature_cover_urls
+
+logger = logging.getLogger('tasks')
+
+# NOTE: Series have covers too
```
ntarocco, line 21: Is this commented because you will do it later?

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+@shared_task(ignore_result=True)
+def update_cover_metadata_isbn(record):
+    """Update record cover metadata isbn."""
+    cover_meta = record.get("cover_metadata", {})
+    if cover_meta.get("isbn"):
+        return
+
+    identifies = record.get("identifies", [])
```
ntarocco, line 43:
```suggestion
    identifiers = record.get("identifiers", [])
```
there is a typo, were you able to test this?

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+    for ident in identifies:
+        if ident["schema"] == "ISBN":
+            record["cover_metadata"]["isbn"] = ident["value"]
+            record.commit()
```
ntarocco, line 47: the idea of the before/after signals is that you can modify the record, but not commit.
The commit will be done by `invenio-records-rest` at the end of the request, otherwise here you will have a second commit.
To be tested.

cds_books/literature/ext.py
```diff
@@ -0,0 +1,30 @@
...
+    """CdsBooksLiterature extension."""
+
+    def __init__(self, app=None):
+        """Literature initialization."""
+        self.register_signals(app)
+
+    @staticmethod
+    def register_signals(app):
+        """Register Literature signals."""
+        before_record_insert.connect(update_cover)
```
ntarocco, line 27: Great! I just find confusing the naming because it will be hard to remember what is the difference. What do you think?
```suggestion
        before_record_insert.connect(add_isbn_to_cover_metadata)
```

cds_books/literature/ext.py
```diff
@@ -0,0 +1,30 @@
...
+        before_record_update.connect(update_cover)
+        after_record_insert.connect(set_cover)
```
ntarocco, line 29:
```suggestion
        after_record_insert.connect(find_existing_cover)
```
Not sure about this, something better probably. But you get my point I guess :)

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+@shared_task(ignore_result=True)
+def set_cover_metadata_isbn(record):
+    """Set record cover metadata isbn with valid cover."""
+
+    def is_valid_response(response):
+        """Validates if the syndetic response returned a cover."""
+        # NOTE: When there is no cover, syndetic returns a 1x1 pixel image
+        if (
+            response.getcode() == 200 and
+            response.get("Content-Length") > MIN_CONTENT_LENGTH
```
ntarocco, line 68: Nice solution! Well thought!
```suggestion
            more_than_1_pixel = response.get("Content-Length") > MIN_CONTENT_LENGTH
```

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+    # 1. try if existing cover_metadata are valid
+    cover_meta = record.get("cover_metadata", {})
+    if cover_meta.get("isbn"):
+        url = build_literature_cover_urls(record)['small']
+        try:
+            resp = urllib.request.urlopen(url)
+            if is_valid_response(resp):
+                return
+        except Exception as e:
+            logger.exception(e)
```
ntarocco, line 82: I am not sure we need this, I will check later.

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # 2. assign the first valid isbn identifier
+    identifies = record.get("identifies", [])
```
ntarocco, line 85:
```suggestion
    identifiers = record.get("identifiers", [])
```

cds_books/literature/tasks.py
```diff
@@ -0,0 +1,97 @@
...
+    for ident in identifies:
+        if ident["schema"] == "ISBN":
+            record["cover_metadata"]["isbn"] = ident["value"]
```
ntarocco, line 88: I am not sure if modifying the record in any case is safe. If there is record.commit() after, it might keep that change, and then we have nasty bugs. Probably it is not true, I would anyway just to be safe to change the record below:
```
if is_valid_response(resp):
    record["cover_metadata"]["isbn"] = ident["value"]
    record.commit()
    db.session.commit()
    return
```

tests/api/test_literature_covers.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: Great thanks! Can you add a test for the task too?

cds_books/literature/ext.py
```diff
@@ -0,0 +1,30 @@
...
+"""Literature module."""
+
+
+from invenio_records.signals import after_record_insert, after_record_update, \
+    before_record_insert, before_record_update
+
+from .tasks import set_cover, update_cover
+
+
+class CdsBooksLiterature(object):
```
ntarocco, line 17: I would go for maybe creating a CdsIls extension instead, a main one.

setup.py
```diff
@@ -44,6 +44,7 @@
             'migration = cds_books.migrator.cli:migration',
         ],
         'invenio_base.apps': [
```
ntarocco, line 46:
```suggestion
        'invenio_base.api_apps': [
```

cds_books/literature/covers.py
kprzerwa: 👍to that

cds_books/literature/__init__.py
topless: I refactored `cds-books`, I checked also just in case to fix ILS, but its bigger than I thought and few things might break.

cds_books/literature/ext.py
topless: Yes that makes sense, I went for `CdsBooks`, to match every other occurrence. We can move the repository and rename everything in one go.

cds_books/literatures/tasks.py
```diff
@@ -0,0 +1,124 @@
...
+    return None
+
+
+@shared_task(ignore_result=True)
+def select_cover_task(record):
+    """Set record cover metadata isbn with valid cover."""
+    if not is_allowed(record):
+        return
+
+    if is_cover_valid(record):
```
ntarocco, line 94:
```suggestion
    cover_metadata = record.get("cover_metadata", {})
    has_already_a_cover = cover_metadata.get("ISBN", False) or cover_metadata.get("ISSN", False)
    if has_already_a_cover and is_cover_valid(record):
```

cds_books/literatures/tasks.py
```diff
@@ -0,0 +1,124 @@
...
+        return
+
+    identifiers = record.get("identifiers", [])
+    if not identifiers:
+        return
+
+    if "cover_metadata" not in record:
```
ntarocco, line 101:
```suggestion
    record.setdefault("cover_metadata", {})
```

cds_books/literatures/tasks.py
```diff
@@ -0,0 +1,124 @@
...
+
+    for ident in identifiers:
+        if ident["scheme"] != "ISBN":
+            continue
+
+        record["cover_metadata"]["isbn"] = ident["value"]
+        if is_cover_valid(record):
+            record.commit()
+            db.session.commit()
+            return
```
ntarocco, line 124: there might be an issue here: if I have many ISSNs, but none of them are valid, and one valid ISBN, what will happen is:
```
cover_metadata: {
    issn: 342423434  # last ISSN that was not valid
    isbn:  342344342   # valid ISBN
}
```
I would avoid changing the record metadata if not needed.

tests/api/test_literature_covers.py
```diff
@@ -0,0 +1,89 @@
...
+    def read(self):
+        return self.resp_data
+
+    def getcode(self):
+        return self.code
+
+
+def test_select_cover_task(app, testdata):
+    """Test to update cover_metadata."""
+    app.config.update(CELERY_TASK_ALWAYS_EAGER=True)
```
ntarocco, line 76: this should be already in the `conftest.py`, otherwise we should add it there and not here.

tests/api/test_literature_covers.py
```diff
@@ -0,0 +1,89 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: Great thanks!

cds_books/config.py
```diff
@@ -172,6 +173,7 @@ def _parse_env_bool(var_name, default=None):
 #: should be set to the correct host and it is strongly recommended to only
 #: route correct hosts to the application.
 APP_ALLOWED_HOSTS = [
+    "127.0.0.1",
```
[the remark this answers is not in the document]
kprzerwa: invenio.cfg would work normally, but it would be a bit painful with pipenv ;/ I don't remember any other way

cds_books/literatures/covers.py
```diff
@@ -0,0 +1,30 @@
...
+    """Decorate literatures with cover urls for all sizes."""
+    client = os.environ.get("SYNDETIC_CLIENT")
+    url = "https://secure.syndetics.com/index.aspx"
+    cover_metadata = metadata.get("cover_metadata", {})
+
+    issn = cover_metadata.get("issn", "")
+    isbn = cover_metadata.get("isbn", "")
+    scheme = 'issn' if issn else 'isbn'
+    scheme_value = issn if issn else isbn
+    return {
```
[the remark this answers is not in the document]
ntarocco: localhost_3000_literature_zx68m-zq625

This has to be fixed in ILS too, see screnshot with no cover. It should have the placeholder instead.

cds_books/config.py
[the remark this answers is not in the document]
topless: @ntarocco sorry for this just slipped in, I use as @kprzerwa states I use the `invenio.cfg`, that was me just being lazy due to multiple times creating the virtual environment.

cds_books/literature/covers.py
```diff
@@ -0,0 +1,88 @@
...
+    if issn or isbn:
+        return {
+            "small": "{url}?client={client}&{sheme}={scheme_value}/SC.gif".format(
+                url=url, client=client, sheme=scheme, scheme_value=scheme_value),
+            "medium": "{url}?client={client}&{sheme}={scheme_value}/MC.gif".format(
+                url=url, client=client, sheme=scheme, scheme_value=scheme_value),
+            "large": "{url}?client={client}&{sheme}={scheme_value}/LC.gif".format(
+                url=url, client=client, sheme=scheme, scheme_value=scheme_value),
+        }
+    return build_placeholder_urls()
```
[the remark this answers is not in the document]
topless: Yes, https://github.com/topless/invenio-app-ils/blob/9eb5d36c7741e1d0706174ad87f8403213b11384/invenio_app_ils/literature/covers_builder.py#L33

PR #113, opened 2020-07-07
migration: schema update fixes

kprzerwa:
* verbose status printing

### DIFF
--- a/cds_books/literature/covers.py
+++ b/cds_books/literature/covers.py
@@ -15,11 +15,12 @@ from invenio_app_ils.literature.covers_builder import build_placeholder_urls
 
 def is_record_with_cover(record):
     """Check if this type of record has cover."""
-    schema = record["$schema"]
-    if schema.endswith("document-v1.0.0.json") or schema.endswith(
-        "series-v1.0.0.json"
-    ):
-        return True
+    if '$schema' in record:
+        schema = record["$schema"]
+        if schema.endswith("document-v1.0.0.json") or schema.endswith(
+            "series-v1.0.0.json"
+        ):
+            return True
     return False
 
 
--- a/cds_books/migrator/api.py
+++ b/cds_books/migrator/api.py
@@ -18,6 +18,8 @@ from flask import current_app
 from invenio_app_ils.documents.api import Document, DocumentIdProvider
 from invenio_app_ils.documents.search import DocumentSearch
 from invenio_app_ils.records_relations.api import RecordRelationsParentChild
+from invenio_app_ils.relations.api import SERIAL_RELATION, \
+    MULTIPART_MONOGRAPH_RELATION
 from invenio_app_ils.series.api import Series, SeriesIdProvider
 from invenio_app_ils.series.search import SeriesSearch
 from invenio_base.app import create_cli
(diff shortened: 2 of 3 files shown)


cds_books/literature/covers.py
```diff
@@ -15,11 +15,12 @@
...
 def is_record_with_cover(record):
     """Check if this type of record has cover."""
-    schema = record["$schema"]
-    if schema.endswith("document-v1.0.0.json") or schema.endswith(
-        "series-v1.0.0.json"
-    ):
-        return True
+    if '$schema' in record:
+        schema = record["$schema"]
+        if schema.endswith("document-v1.0.0.json") or schema.endswith(
```
topless, line 20: question: we are using this `endswith` a lot for matching the right schema comparisons, my concern is that these will fail if we change the schema version.

cds_books/migrator/api.py
```diff
@@ -150,9 +162,11 @@ def get_multipart_by_legacy_recid(recid):
         ]
     )
     result = search.execute()
-    if result.hits.total < 1:
-        raise MultipartMigrationError(
-            'no multipart found with legacy recid {}'.format(recid))
+    if not result.hits or result.hits.total < 1:
+        click.secho('no multipart found with legacy recid {}'.format(recid),
+                    fg='red')
+        # raise MultipartMigrationError(
```
topless, line 168: nit: Is this used? is it a test? no commented code.

cds_books/migrator/api.py
```diff
@@ -186,18 +201,18 @@ def create_multipart_volumes(pid, multipart_legacy_recid, migration_volumes):
     first_volume = next(volume_numbers)
     first = Document.get_record_by_pid(pid)
     if 'title' in volumes[first_volume]:
-        first['title']['title'] = volumes[first_volume]['title']
+        first['title'] = volumes[first_volume]['title']
         first['volume'] = first_volume
     first['_migration']['multipart_legacy_recid'] = multipart_legacy_recid
-    if 'legacy_recid' in first:
-        del first['legacy_recid']
+    # if 'legacy_recid' in first:
```
topless, line 207: no commented code

cds_books/migrator/api.py
```diff
@@ -339,8 +361,8 @@ def validate_serial_relation(serial, recids):
     search = SeriesSearch().filter('term', mode_of_issuance='SERIAL')
     for serial_hit in search.scan():
         # Store titles and check for duplicates
-        if 'title' in serial_hit and 'title' in serial_hit.title:
-            title = serial_hit.title.title
+        if 'title' in serial_hit:
```
topless, line 364: nit: `title = serial_hit.get('title', '')`

cds_books/migrator/records.py
```diff
@@ -233,13 +240,20 @@ def create_record(cls, dump):
...
-        db.session.commit()
+        try:
+            document = Document.create(record.model.json, record_uuid)
+            document.commit()
+            db.session.commit()
+
+            return document
+        except IlsValidationError as e:
+            click.secho(e.original_exception, fg='red')
+            # TODO uncomment when data cleaner
```
topless, line 257: nit: there are a lot of these comments, an some of them are uncommented already.

cds_books/literature/covers.py
kprzerwa: there were discussions on checking the record type instead, although we should decide together about this, and this is only a fix for covers being calculated for intermediate record revision provided by the migration, which does not have `$schema` yet, and it was throwing errors. Also it would happen for any other record as it is a signal in invenio-rest

cds_books/migrator/api.py
kprzerwa: This is not a final form and I left the comment to remember to change it when we have cleaner data. Otherwise I can't run the scripts and test relations on a bigger set of input because everything fails. Are there better ideas how to remember about it rather that leaving commented code?

cds_books/migrator/api.py
kprzerwa: same explanation as the previous comment

PR #114, opened 2020-07-07
home: changes bkg image and logo

ntarocco:
* changes bkg image in homepage
* changes favicon and logo
* removes unused files

cds_books_homepage

cds_books_footer


topless:
nit: @ntarocco can I suggest to try to add some `opacity` in the background image, so our controls might stand out a bit more? I am not sure about the result, just a try.

kprzerwa:
> nit: @ntarocco can I suggest to try to add some `opacity` in the background image, so our controls might stand out a bit more? I am not sure about the result, just a try.

there is already a dark filter, we could increase the darkness eventually

ntarocco:
> > nit: @ntarocco can I suggest to try to add some `opacity` in the background image, so our controls might stand out a bit more? I am not sure about the result, just a try.
> 
> there is already a dark filter, we could increase the darkness eventually

@topless @kprzerwa as discussed, improving CSS filters to the background images will require to change the CSS of ILS. I would merge as it is this one, and apply changes to ILS.

PR #131, opened 2020-08-07
Add delete_users in ldap 

### LINKED ISSUE #53: ldap: Implement deletion of users when syncing with ldap
Requires: https://github.com/CERNDocumentServer/cds-books/issues/59
Requires: https://github.com/inveniosoftware/invenio-app-ils/issues/925

In CERN LDAP, users leaving CERN are:
- blocked after 2 months
- deleted after 6 months

Ask @zzacharo: how do we know if the user is blocked? Is there any field in LDAP marking that?

@zzacharo Answer: Yes there is a field indicating that but after discussion, a user can be marked as blocked also for example because they forgot their password. So, it's not an indicator that a person left CERN, especially because you cannot retrieve from ldap the reason why they are blocked. I think, the rule of "deleted after 6 months" i.e the user cannot be retrieved from ldap is enough for us to clean up our database. 

### Tasks
On each LDAP sync:
- when the user is deleted i.e not found in ldap:
  - [x] if user has current ongoing loans, send an email to the library with the link to the patron and info (because it will be anonymosed in the system just after)
  - [x] Anonymise all patron's loan history/activity, acquisitions, ILLs, Document Requests, etc... (should be done by https://github.com/inveniosoftware/invenio-app-ils/issues/925, just a matter to call that function)

KonstantinaStoikou:
Closes #53 

- Call `anonymize_user` from [this PR](https://github.com/inveniosoftware/invenio-app-ils/pull/927)
- In case user has ongoing loans call `send_active_loans_mail` from [this PR](https://github.com/inveniosoftware/invenio-app-ils/pull/928)
- Override footer email template from `invenio-app-ils`


tests/ldap/conftest.py
```diff
@@ -9,11 +9,38 @@
...
+    #         "invenio_app_ils_tests",
+    #         __name__,
+    #         template_folder="../cds_books/circulation/templates",
+    #     )
+    # )
+    # add extra test templates to the search app blueprint, to fake the
+    # existence of `invenio-theme` base templates.
+    test_templates_path = os.path.join(os.path.dirname(__file__), "templates")
+    print("PATH:", test_templates_path)
+    test_templates_path = "/home/konstantina/Documents/Work/cds-books/cds_books/circulation/templates"
```
jrcastro2, line 40: This seems to be for local testing, or am I missing something?

KonstantinaStoikou: Yes you are right, I forgot it

PR #133, opened 2020-08-11
ldap: users synchronization enhancements

### LINKED ISSUE #59: ldap: enhance tasks
The sync of CERN users with LDAP should run every day and it can succeed or fail. Currently it is only a CLI command executed when needed.
It would be nice to display the task executions in the Invenio Admin interface, so that admins can immediately check status and see errors.

### Tasks

- [x] refactor the CLI command `sync_users` to be a dedicated function that can be called on demand by the existing CLI command or a cron task
- [x] create a cron task to run every night and execute this sync
- [x] refactor the code to print very clear console messages in a structure way (JSON logging) so that such logs can be eventually sent to Elasticsearch and we can run queries on it (see example in InvenioILS `ReferencedRecordsIndexer`). Ensure that exceptions are sent to `Sentry` (normally it is automatically happening when using console.error or console.exception, to be checked), please test it
- [x] create a new db table to store information of each task execution, one row per task exec. Each time the function is run, by a celery task or manually via CLI, it should report at least (db columns):
  - agent (or similar): CLI or celery cron (choose appropriate naming)
  - task id, if celery cron
  - status: RUNNING/SUCCEEDED/FAILED
  - message: might be a report message if running, e.g. fetching users, and the error message (stacktrace or not?) if error
  - start datetime
  - end datetime
  - how many CERN users fetched from LDAP
  - how many users updated in ILS (because some fields changed)
  - how many users deleted from ILS
  - how many users added in ILS
  - any other field?
- [x] create an admin panel in `invenio-admin` to display these rows. Add a button to trigger a new celery task on demand.

FlorianCassayre:
* refactored the ldap user synchronization function
* new database model to store synchronization task logs
* includes an invenio admin view
* json structured messages in logger
* closes #59

image

image

### DIFF
--- a/cds_books/config.py
+++ b/cds_books/config.py
@@ -16,7 +16,10 @@ You overwrite and set instance-specific configuration by either:
 from __future__ import absolute_import, print_function
 
 import os
+from datetime import timedelta
 
+from invenio_app_ils.config import \
+    CELERY_BEAT_SCHEDULE as ILS_CELERY_BEAT_SCHEDULE
 from invenio_app_ils.config import RECORDS_REST_ENDPOINTS
 from invenio_app_ils.patrons.api import PATRON_PID_TYPE
 from invenio_records_rest.schemas.fields import SanitizedUnicode
@@ -136,6 +139,14 @@ BROKER_URL = "amqp://guest:guest@localhost:5672/"
 CELERY_BROKER_URL = "amqp://guest:guest@localhost:5672/"
 #: URL of backend for result storage (default is Redis).
 CELERY_RESULT_BACKEND = "redis://localhost:6379/2"
+#: Scheduled tasks configuration (aka cronjobs).
+CELERY_BEAT_SCHEDULE = {
+    **ILS_CELERY_BEAT_SCHEDULE,  # Parent config
+    "synchronize_users": {
+        "task": "cds_books.ldap.tasks.synchronize_users_task",
+        "schedule": timedelta(days=1),
+    },
+}
 
 # Database
 # ========
--- a/cds_books/ldap/admin.py
+++ b/cds_books/ldap/admin.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# CDS Books is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Admin panel for ldap synchronizations."""
+
+import flask
+from flask import Blueprint, current_app, flash, redirect, request
+from flask_admin.base import expose
+from flask_admin.contrib.sqla import ModelView
+from flask_admin.helpers import get_redirect_target
+from flask_babelex import gettext as _
+from invenio_db import db
+
+from cds_books.ldap.models import LdapSynchronizationLog
+from cds_books.ldap.tasks import synchronize_users_task
+
+
+class LdapSynchronizationLogModelView(ModelView):
+    """Invenio admin view for ldap users synchronization."""
+
+    # Entries are read-only
+    can_create = True
+    can_edit = False
+    can_delete = False
+
+    can_view_details = True
+
+    form_excluded_columns = LdapSynchronizationLog.__table__.columns
+    column_default_sort = (LdapSynchronizationLog.start_time, True)
+
+    @expose('/new/', methods=('GET', 'POST'))
+    def create_view(self):
+        """Override the creation form and replace it by an action button."""
+        if flask.request.method == 'POST':
+            try:
+                self.start_task()
+                flash('The task was successfully started.', 'success')
+            except Exception as e:
+                current_app.logger.exception(e)
+                flash('An error occurred while starting the task.', 'error')
+            return redirect(request.path)  # Redirect after POST
+
+        return_url = get_redirect_target() or self.get_url('.index_view')
+
+        return self.render('cds_books_admin/create_task.html',
+                           return_url=return_url
+                           )
+
+    @staticmethod
+    def start_task():
+        """Start the task as requested by the user."""
+        synchronize_users_task.apply_async()
+
+
+blueprint = Blueprint(
+    'cds_books_admin',
+    __name__,
+    template_folder='templates',
+    static_folder='static',
+)
+
+ldap_sync = {
+    'model': LdapSynchronizationLog,
+    'modelview': LdapSynchronizationLogModelView,
+    'name': 'Ldap Synchronization',
+    'category': _('User Management')
+}
+
+__all__ = (
+    'ldap_sync',
+)
--- a/cds_books/ldap/api.py
+++ b/cds_books/ldap/api.py
@@ -7,13 +7,18 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """CDS Books ldap API."""
+import json
+import uuid
 
 import ldap
 from flask import current_app
 from invenio_accounts.models import User
+from invenio_app_ils.anonymization import anonymize_patron_data
+from invenio_app_ils.circulation.tasks import send_active_loans_mail
 from invenio_db import db
 from invenio_oauthclient.models import RemoteAccount, UserIdentity
 from invenio_userprofiles.models import UserProfile
+from sqlalchemy.orm.exc import NoResultFound
 
 
 class LdapClient(object):
@@ -32,12 +37,12 @@ class LdapClient(object):
     """
 
     LDAP_USER_RESP_FIELDS = [
-        'mail',
-        'displayName',
-        'department',
-        'cernAccountType',
-        'employeeID',
-        'uidNumber'
+        "mail",
+        "displayName",
+        "department",
+        "cernAccountType",
+        "employeeID",
+        "uidNumber",
     ]
 
     def __init__(self, ldap_url):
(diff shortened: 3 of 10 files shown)


FlorianCassayre:
Important notes:
We did not manage to verify that the Celery task was scheduled correctly.
Moreover it's unclear if by redefining the `CELERY_BEAT_SCHEDULE` config list field it would override the one defined in ILS.

cds_books/config.py
```diff
@@ -128,22 +128,31 @@ def _parse_env_bool(var_name, default=None):
...
+BROKER_URL = "amqp://guest:guest@localhost:5672/"
 #: URL of message broker for Celery (default is RabbitMQ).
-CELERY_BROKER_URL = 'amqp://guest:guest@localhost:5672/'
+CELERY_BROKER_URL = "amqp://guest:guest@localhost:5672/"
 #: URL of backend for result storage (default is Redis).
-CELERY_RESULT_BACKEND = 'redis://localhost:6379/2'
+CELERY_RESULT_BACKEND = "redis://localhost:6379/2"
+#: Scheduled tasks configuration (aka cronjobs).
+CELERY_BEAT_SCHEDULE = {
+    **CELERY_BEAT_SCHEDULE,  # Parent config
```
zzacharo, line 138: minor: Maybe rename it in `ILS_CELERY_BEAT_SCHEDULE` just to be clearer?

cds_books/ldap/admin.py
```diff
@@ -0,0 +1,70 @@
...
+
+    can_create = True
+    can_edit = False
+    can_delete = False
+    can_view_details = True
+    form_excluded_columns = LdapSynchronizationLog.__table__.columns
+    column_default_sort = (LdapSynchronizationLog.start_time, True)
+
+    @expose('/new/', methods=('GET', 'POST'))
+    def create_view(self):
```
zzacharo, line 33: @FlorianCassayre I think we might have a problem with permissions here. Have you checked that only authenticated admin users can post on this one?

tests/ldap/test_ldap_cli.py
```diff
@@ -51,3 +54,26 @@ def test_import_ldap_users(app):
...
+    # Basic insertion
+    log = LdapSynchronizationLog.create_cli()
+    assert find()
+    assert find().status == TaskStatus.RUNNING and find().agent == Agent.CLI
+    find().query.delete()
+    assert not find()
+
+    # Change state
+    log = LdapSynchronizationLog.create_celery('1')
+    assert find().status == TaskStatus.RUNNING and find().agent == Agent.CELERY
```
zzacharo, line 73: same here :)

cds_books/ldap/cli.py
```diff
@@ -81,51 +44,18 @@ def import_users():
...
-
-    click.secho(
-        "--- Finished in %s seconds ---" % (time.time() - start_time),
-        fg="green"
-    )
+def ldap_sync_users():
+    """Add entry to database with info about this synchronization."""
+    log = LdapSynchronizationLog.create_cli()
+    result = sync_users()
+    print(result)
```
zzacharo, line 57: Is this print intended? If yes, why not using the `click.secho()` or a logger?

cds_books/ldap/api.py
```diff
@@ -175,24 +190,183 @@ def _commit_user(user_data):
...
+    )
+    structured_msg_str = json.dumps(structured_msg, sort_keys=True)
+    current_app.logger.info(structured_msg_str)
+
+
+def _log_exception(exception):
+    current_app.logger.exception(exception)
+
+
+def sync_users():
```
zzacharo, line 298: Same thought as above about if this should be defined here or in `cli.py`! Maybe also @kprzerwa @KonstantinaStoikou @jrcastro2 can share their thoughts!

cds_books/ldap/cli.py
```diff
@@ -81,51 +44,18 @@ def import_users():
...
+    if result is None:
+        log.transition_fails()
```
zzacharo, line 59: do we have any useful information that we can pass in the case that the task fails? Similar to the case it succeeds below?

cds_books/ldap/tasks.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@shared_task
+def synchronize_users_task():
+    """Add entry to database with info about this synchronization."""
+    log = LdapSynchronizationLog.create_celery(
+        synchronize_users_task.request.id
+    )
+    result = sync_users()
+    print(result)
```
kprzerwa, line 27: print 👀

cds_books/ldap/admin.py
```diff
@@ -0,0 +1,70 @@
...
+from invenio_db import db
+
+from cds_books.ldap.models import LdapSynchronizationLog
+from cds_books.ldap.tasks import synchronize_users_task
+
+
+class LdapSynchronizationLogModelView(ModelView):
+    """Invenio admin view for ldap users synchronization."""
+
+    can_create = True
```
kprzerwa, line 25: how are these attributes used?

cds_books/config.py
FlorianCassayre: This was intended, in order to emphasize that the field 'kind' is the same and has to come from the parent. As a matter of taste I prefer it the current way, but I can change it if needed.

cds_books/ldap/admin.py
FlorianCassayre: I recall trying to access this very page as an anonymous user, and the system would redirect me to the login form. I'm assuming it's handled internally by `flask-admin`, but I'll double check.

cds_books/ldap/api.py
```diff
@@ -175,24 +190,183 @@ def _commit_user(user_data):
...
-                                    user_id, ldap_user))
+                client_id=client_id,
+                **self.import_remote_account(user_id, ldap_user)
+            )
             db.session.add(remote_account)
 
         db.session.commit()
+
+
+def index_ldap_users():
```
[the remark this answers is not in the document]
FlorianCassayre: This function is used in both `cli.py` and `api.py` (by `sync_users`, which is itself used in `cli.py` and `tasks.py`), that was the motivation behind the refactoring.

cds_books/ldap/api.py
```diff
@@ -175,24 +190,183 @@ def _commit_user(user_data):
...
+    cli = create_cli()
+    runner = current_app.test_cli_runner()
+    command = "ils patrons index"
+    click.secho("ils {}...".format(command), fg="green")
+    runner.invoke(cli, command, catch_exceptions=True)
+    _log_info("command_executed", dict(command=command))
+
+
+def import_ldap_users(ldap_users):
+    """Import ldap users in db."""
```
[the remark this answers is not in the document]
FlorianCassayre: [ditto](https://github.com/CERNDocumentServer/cds-books/pull/133/files/142527fb986e67613603676091fdd4fd4b34cf7a#r469996478)

cds_books/ldap/cli.py
FlorianCassayre: I think it is an old print-debug that shall be removed safely (@KonstantinaStoikou).

cds_books/ldap/cli.py
FlorianCassayre: We could return the exception type & message as a string; but we agreed that exceptions would be rare and the message not necessarily useful to the end user. Moreover, the logger already records the stacktrace via Sentry.

cds_books/ldap/api.py
```diff
@@ -47,14 +51,15 @@ def __init__(self, ldap_url):
     def get_primary_accounts(self):
         """Retrieve all primary accounts from ldap."""
         page_control = ldap.controls.SimplePagedResultsControl(
-                            True, size=1000, cookie='')
+            True, size=1000, cookie=""
+        )
 
         response = self.ldap.search_ext(
-            'OU=Users,OU=Organic Units,DC=cern,DC=ch',
+            "OU=Users,OU=Organic Units,DC=cern,DC=ch",
```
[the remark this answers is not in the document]
FlorianCassayre: This is snippet was already there (I believe Konstantina's linter replaces single quotes by doubles). Regardless, we can of course add a comment about it.

cds_books/ldap/admin.py
FlorianCassayre: Those come from `flask-admin`, which is itself provided by `invenio-admin` (see [the documentation](https://flask-admin.readthedocs.io/en/latest/introduction#modelview-configuration-attributes)).

cds_books/ldap/admin.py
FlorianCassayre: I confirm that, as for the other pages, it:
- redirects to login page when not logged-in
- returns `403` when logged-in as a patron 
image

cds_books/ldap/cli.py
FlorianCassayre: I added a new `message` field taking the value `${exception_type}: ${exception_message}` if an exception is caught or `NULL` otherwise. The full stacktrace is still available on the above-mentioned service.
image

cds_books/ldap/api.py
kprzerwa: cli would be more intuitive for us I think

cds_books/ldap/api.py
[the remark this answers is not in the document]
zzacharo: This is ldap specific which I copied from another implementation...The parameters `OU, DC` are ldap specific...I know that `OU` means `organizational unit` but I don't recall the other one...

cds_books/ldap/api.py
```diff
@@ -65,32 +70,38 @@ def get_primary_accounts(self):
...
-            controls = [control for control in serverctrls
-                        if control.controlType == ldap_page_control_type]
+            controls = [
+                control
+                for control in serverctrls
+                if control.controlType == ldap_page_control_type
+            ]
             if not controls:
-                print('The server ignores RFC 2696 control')
+                print("The server ignores RFC 2696 control")
```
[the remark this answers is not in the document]
zzacharo: This was also me :P Again, didn't dig more into this...but I think we can remove this check ;) I would say is specific to how the ldap server works which in our case doesn't fail, so there is no point to check as ldap won't change soon!

kprzerwa:
🚀

PR #138, opened 2020-08-31
rename: Rename extension fields names

### LINKED ISSUE #121: rename extension fields names
Standard_CERN_status - > standard_status or standard_review
accelerators_experiments -> unit/organisation_unit/related_unit

jrcastro2:
* standard_CERN_status -> standard_review
* accelerators_experiments -> unit
* closes #121

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -267,11 +267,11 @@ ILS_LITERATURE_COVER_URLS_BUILDER = build_syndetic_cover_urls
 # Namespaces for fields added to the metadata schema
 ILS_RECORDS_METADATA_NAMESPACES = {
     "document": {
-        "accelerator_experiments": {
-            "@context": "https://cern.ch/accelerator_experiments/terms"
+        "unit": {
+            "@context": "https://cern.ch/unit/terms"
         },
-        "standard_CERN_status": {
-            "@context": "https://cern.ch/standard_CERN_status/terms"
+        "standard_review": {
+            "@context": "https://cern.ch/standard_review/terms"
         },
     }
 }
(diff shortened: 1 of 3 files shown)


cds_ils/config.py
```diff
@@ -267,63 +267,63 @@ def _parse_env_bool(var_name, default=None):
...
             "elasticsearch": "keyword",
             "marshmallow": SanitizedUnicode(),
         },
-        "accelerator_experiments:study": {
+        "unit:study": {
             "elasticsearch": "keyword",
             "marshmallow": SanitizedUnicode(),
         },
-        "standard_CERN_status:CERN_applicability": {
+        "standard_review:CERN_applicability": {
```
jrcastro2, line 310: In this PR <https://github.com/CERNDocumentServer/cds-dojson/pull/266/files#diff-8c22f650d9678abd75eb860a77a826aaR275> there is a `standard_review:applicability`, instead of `standard_review:CERN_applicability`, and on the excel https://docs.google.com/spreadsheets/d/1OHMDiMfBD1Aam-kteIuBL9tEPK-C-Jevw-NcmGGxRvY/edit#gid=1257315908 it's also `applicability` (search for standard_status.applicability). I guess that it should change to `standard_review:applicability`. Since it's not part of the task can someone confirm it?

FlorianCassayre: I would say so (they probably forgot to mention it). To be confirmed.

topless: I would go for `standard_review:applicability`, since we are on CERN overlay I believe we don't have to mention it in every field.

topless (approved):

cds_ils/config.py
jrcastro2: I completely agree, thanks!

ui/src/config.js
```diff
@@ -19,63 +19,63 @@ export const config = {
...
           default: '',
         },
-        'accelerator_experiments:institution': {
+        'unit:institution': {
           label: 'Institution',
           type: 'string',
           default: '',
         },
-        'accelerator_experiments:legacy_name': {
+        'unit:legacy_name': {
```
kprzerwa, line 42: shouldn't this be removed in #139 ?

kprzerwa (approved):

PR #143, opened 2020-09-09
global: fix failing API calls due to frame_options

kprzerwa:


cds_ils/config.py
```diff
@@ -182,6 +183,9 @@ def _parse_env_bool(var_name, default=None):
     os.environ.get("HOSTNAME", ""),  # fix disallowed host error during /ping
 ]
 
+APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
```
ntarocco, line 186: Please remove this line:
```suggestion
APP_DEFAULT_SECURE_HEADERS["content_security_policy"] = {}
```
We should actually remove it from ILS too, and set the right CSP headers!!!

cds_ils/config.py
```diff
@@ -182,6 +183,9 @@ def _parse_env_bool(var_name, default=None):
...
+APP_DEFAULT_SECURE_HEADERS["frame_options"] = None
```
ntarocco, line 187: Please remove this line. @zzacharo I have fixed the issue normally, see [here](https://gitlab.cern.ch/AIGROUP-cds-admin/cds-books-openshift/-/commit/30f5341398a207c22b2e594251d96d42891d8565).
The issue is that Nginx by default does not pass headers with `_`, see doc [here](http://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers)

ntarocco (approved):

PR #144, opened 2020-09-14
covers: update cover when identifier gets updated

### LINKED ISSUE #142: covers: when changing a valid identifier with another valid one, cover remains the same
When you change an identifier with a valid cover with another one that also has a valid cover, the cover is not updated to the new identifier.

Also when removing all identifiers, cover should be the placeholder one.

KonstantinaStoikou:
Closes #142


KonstantinaStoikou:
WIP: Refactoring to a more simple solution.

ntarocco:
@KonstantinaStoikou let's implement your idea. Let's re-implement it and abandon the `preemptively` guessing, it is too messy...
If `cover_metadata` is empty, then the task should run and try to find the valid cover.
If `cover_metadata` is not empty, then the task should not run, because the cover was already set at some point. 

1. remove the `before_record_insert` signal.
2. in the `before_record_update`, let's remove any previous cover, every single time, even if it was valid and it will be set to the same:
```
    if not should_record_have_cover(record):
        return
    else:
        record["cover_metadata"] = {}
```
3. in the `after_record_insert/after_record_update`, we check that if there is nothing, we test the ISBN/ISSN:
```
def pick_identifier_with_cover(sender, *args, **kwargs):
    """Triggers async task to set cover metadata of the record."""
    record = kwargs.get("record", {})
    if should_record_have_cover(record) and record.get("cover_metadata", {}):
        # trigger task only if "cover_metadata" is empty
        pick_identifier_with_cover_task.apply_async((record,))

@shared_task(ignore_result=True)
def pick_identifier_with_cover_task(record):
    """Set an valid cover identifier to cover metadata, else an empty dict."""
    if not should_record_have_cover(record) or not record.get("cover_metadata", {}):
        return

    identifiers = record.get("identifiers", [])
    if not identifiers:
        return

    ... pick the good identifier if any and save, otherwise do not set record["cover_metadata"] and do not save ...
```

Let me know!

KonstantinaStoikou:
I removed completely the `before_record` signals because they created infinite loops.

kprzerwa (approved):

PR #146, opened 2020-09-15
ui: fix link help menu

ntarocco:


ui/src/overridden/components/Footer/Footer.js
```diff
@@ -59,16 +59,16 @@ export const Footer = ({ ...props }) => {
               <List>
                 <List.Item>+41 22 767 24 35</List.Item>
```
topless, line 60: nit: should we also make the number part of config?

ui/src/overridden/components/Menu/RightMenuItem.js
```diff
@@ -1,46 +1,55 @@
...
-    <Dropdown.Divider />
-    <Dropdown.Item as={Link} to={FrontSiteRoutes.documentRequestForm}>
-      Request new literature
-    </Dropdown.Item>
-  </Dropdown.Menu>
-);
+const DropdownItems = () => {
+  return (
+    <>
+      <Dropdown.Item as="a" href="https://library.web.cern.ch/resources/remote">
```
topless, line 13: nit: same here to make it part of the config, domains change etc..

topless (approved):

ui/src/overridden/components/Footer/Footer.js
ntarocco: The configuration is currently hardcoded in the code. The only advantage of having that config at the moment is to avoid duplications.
I would say that it make sense only if we need to re-use.

ui/src/overridden/components/Menu/RightMenuItem.js
ntarocco: See my comment above. A change of its value will require a code change + re-deployment anyway, at the moment.

PR #156, opened 2020-09-30
migrator: ingest loans and user extra data

### LINKED ISSUE #136: migration: add scripts to migrate loans
decisions: only ongoing or the history too?

you need: ingested items, patrons from LDAP

kprzerwa:
closes #136
the screenshots available in the Weekly recap, I didn't add them because of user data protection ;)

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -24,7 +24,7 @@ from invenio_app_ils.config import RECORDS_REST_ENDPOINTS
 from invenio_app_ils.patrons.api import PATRON_PID_TYPE
 from invenio_oauthclient.contrib import cern_openid
 from invenio_records_rest.schemas.fields import SanitizedUnicode
-from marshmallow.fields import Bool
+from marshmallow.fields import Bool, List
 
 from .literature.covers import build_cover_urls
 from .patrons.api import Patron
@@ -249,13 +249,13 @@ OAUTH_REMOTE_REST_APP["logout_url"] = os.environ.get(
     "protocol/openid-connect/logout",
 )
 OAUTH_REMOTE_REST_APP["authorized_redirect_url"] = (
-    os.environ.get("SPA_HOST", "http://localhost:3000") + "/login"
+    os.environ.get("SPA_HOST", "http://127.0.0.1:3000") + "/login"
 )
 OAUTH_REMOTE_REST_APP["disconnect_redirect_url"] = os.environ.get(
-    "SPA_HOST", "http://localhost:3000"
+    "SPA_HOST", "http://127.0.0.1:3000"
 )
 OAUTH_REMOTE_REST_APP["error_redirect_url"] = (
-    os.environ.get("SPA_HOST", "http://localhost:3000") + "/login"
+    os.environ.get("SPA_HOST", "http://127.0.0.1:3000") + "/login"
 )
 OAUTH_REMOTE_REST_APP["params"].update(
     dict(
@@ -284,7 +284,7 @@ CERN_APP_OPENID_CREDENTIALS = dict(
         "OAUTH_CERN_OPENID_CLIENT_ID", "localhost-cds-ils"
     ),
     consumer_secret=os.environ.get(
-        "OAUTH_CERN_OPENID_CLIENT_SECRET", "<change me>"
+        "OAUTH_CERN_OPENID_CLIENT_SECRET", "84fc8be1-12a3-47f7-891e-ba01f158559e"
     ),
 )
 USERPROFILES_EXTEND_SECURITY_FORMS = True
@@ -354,7 +354,7 @@ ILS_RECORDS_METADATA_EXTENSIONS = {
     "document": {
         "unit:accelerator": {
             "elasticsearch": "keyword",
-            "marshmallow": SanitizedUnicode(),
+            "marshmallow": List(SanitizedUnicode()),
         },
         "unit:curated_relation": {
             "elasticsearch": "boolean",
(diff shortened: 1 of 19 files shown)


cds_ils/config.py
```diff
@@ -283,7 +283,7 @@ def _parse_env_bool(var_name, default=None):
         "OAUTH_CERN_OPENID_CLIENT_ID", "localhost-cds-ils"
     ),
     consumer_secret=os.environ.get(
-        "OAUTH_CERN_OPENID_CLIENT_SECRET", "<change me>"
+        "OAUTH_CERN_OPENID_CLIENT_SECRET", "84fc8be1-12a3-47f7-891e-ba01f158559e"
```
ntarocco: remove even if it just a test secret?

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
                     clean_item_record(record)
-                except ItemMigrationError:
+                except ItemMigrationError as e:
+                    click.secho(str(e), fg="red")
                     continue
-                record = import_record(
-                    record, model, provider, legacy_id_key="barcode"
+                try:
+                    # check if the item already there
+                    item = get_item_by_barcode(record["barcode"])
```
ntarocco, line 289: Since the legacy barcode should be unique, if you prefer you can mint an extra PID for each item with the barcode so the search is a primary key lookup instead of ES.

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
+                        continue
+                except ItemMigrationError:
+                    record = import_record(
+                        record, model, provider, legacy_id_key="barcode"
+                    )
+                try:
+                    # without this script is very slow
+                    db.session.commit()
+                except:
+                    db.session.rollback()
```
ntarocco, line 306: log exception?

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
+                    "User {0}({1}) not synced via LDAP".format(
+                        record["id"], record["email"]
+                    ),
+                    fg="red",
+                )
+                continue
+                # todo uncomment when more data
+                # raise UserMigrationError
+            else:
+                client_id = current_app.config["CERN_APP_OPENID_CREDENTIALS"][
```
ntarocco, line 331: does it make sense to refactor the `import_users` function (from LDAP module), and re-use it here? If we fix a bug with the user creation/import from LDAP, we need to remember to fix it here too.

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
+                        record["legacy_id"]
+                    ),
+                    fg="blue",
+                )
+
+            # create a loan
+            (
+                default_location_pid_value,
+                _,
+            ) = current_app_ils.get_default_location_pid
```
ntarocco, line 392: better to put this outside the for loop since it won't change.
Now that I am thinking to this, maybe we want to hardcode the PID in this script somewhere (or pass as param/CLI arg): first we create locations, then we know the CERN library location PID and we pass it.

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
+
+            if record["status"] == "on loan":
+                loan_dict = dict(
+                    patron_pid=str(patron_pid),
+                    transaction_location_pid=default_location_pid_value,
+                    transaction_user_pid=str(SystemAgent.id),
+                    document_pid=document_pid,
+                    item_pid={
```
ntarocco, line 400: what about just:
```suggestion
                    item_pid=item.pid
```
?

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
+                extra_data = account.extra_data
+                # add legacy_id information
+                account.extra_data.update(legacy_id=record["id"], **extra_data)
+                db.session.add(account)
+                patron = Patron(user.id)
+                PatronIndexer().index(patron)
+        db.session.commit()
+
+
+def import_loans_from_json(dump_file):
```
ntarocco, line 346: I think we are missing the loan requests? Table `crcLOANREQUEST`

cds_ils/migrator/api.py
```diff
@@ -250,15 +264,232 @@ def import_items_from_json(dump_file, include, rectype="item"):
...
+def get_item_by_barcode(barcode):
+    """Retrieve item object by barcode."""
+    search = ItemSearch().query(
+        "bool",
+        filter=[
+            Q("term", barcode=barcode),
+        ],
+    )
+    result = search.execute()
+    hits_total = result.hits.total if lt_es7 else result.hits.total.value
```
ntarocco, line 452: you can avoid this if you prefer, we will go with ES7 only

cds_ils/migrator/api.py
```diff
@@ -16,28 +16,37 @@
 from elasticsearch import VERSION as ES_VERSION
```
ntarocco, line 17: Not for this PR: I would re-organize the code because it looks like that it is growing... e.g. class Patron with get and import methods, etc...

cds_ils/migrator/api.py
kprzerwa: I'm working on logging in the next PR

cds_ils/migrator/api.py
kprzerwa: I'm not 100% sure if all the barcodes are unique. but it would speed up the process for sure.

cds_ils/migrator/api.py
kprzerwa: actually I'm doing it in this way, because I rely on the users already existing in the db, synced from LDAP and I am adding only the legacy ID. I don't think it should depend on each other, since these data is from 2 "different" sources

cds_ils/migrator/api.py
kprzerwa: yes, I planned it after files

cds_ils/migrator/api.py
kprzerwa: I started already :)

ntarocco (approved):

PR #161, opened 2020-10-06
Migration files

kprzerwa:
* migrating files stored in the `_migration`
* logging per record type to a file


cds_ils/migrator/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def get_document_by_legacy_recid(legacy_recid):
+    """Search documents by its legacy recid."""
+    search = DocumentSearch().query(
+        "bool", filter=[Q("term", legacy_recid=legacy_recid)]
+    )
+    result = search.execute()
+    hits_total = result.hits.total.value
+    if not result.hits or hits_total < 1:
```
ntarocco, line 96: very minor: it should be enough to check for the total normally

cds_ils/migrator/eitems/api.py
```diff
@@ -0,0 +1,93 @@
...
+    results = get_all_documents_with_files()
+    click.echo(
+        "Found {} documents with files.".format(results.hits.total.value)
+    )
+    for hit in results:
+
+        # make sure the document is in DB not only ES
+        document = Document.get_record_by_pid(hit.pid)
+        click.echo("Processing document {}...".format(document["pid"]))
+        for file_link in document["_migration"]["eitems_file_links"]:
```
ntarocco, line 71: will the `_migration` obj stay forever in the record? If yes, just make sure that we are not putting anything "bad" in it

cds_ils/migrator/eitems/api.py
```diff
@@ -0,0 +1,93 @@
...
+
+            click.echo("File: {}".format(file_link))
+            eitem, bucket = create_eitem_with_bucket_for_document(
```
ntarocco, line 74: what happens if the download fails? I would maybe first download the file, then add it there?
can we get the checksum somehow to be sure that we have a file that is intact?

cds_ils/migrator/eitems/api.py
```diff
@@ -0,0 +1,93 @@
...
+                document["pid"]
+            )
+
+            # get filename
+            file_name = file_link["description"]
+            if not file_name:
+                file_name = file_link.split("/")[-1]
```
ntarocco, line 81: would it be safer to do something like [this](https://stackoverflow.com/a/18727481)?

ntarocco (approved):
nice refactoring!

cds_ils/migrator/eitems/api.py
kprzerwa: no I plan to clean it once everything is migrated, but for now it didn't make sense, because there will be still multiparts handled in this object

cds_ils/migrator/eitems/api.py
kprzerwa: I was thinking about the checksum, I can add it

cds_ils/migrator/eitems/api.py
kprzerwa: 👍

cds_ils/migrator/documents/api.py
kprzerwa: true, leftovers from before I will clean it

PR #163, opened 2020-10-07
Migration external links

kprzerwa:


cds_ils/migrator/documents/api.py
```diff
@@ -118,11 +118,52 @@ def get_document_by_legacy_recid(legacy_recid):
...
+def get_documents_with_external_eitems():
+    """Return documents with eitems from external providers to be migrated."""
+    search = DocumentSearch().filter(
+        "bool",
+        filter=[
+            Q("term", _migration__eitems_has_external=True),
+        ],
+    )
+    total = search.count()
+    search = search[0:total]
```
ntarocco, line 168: Maybe better?
```suggestion
    return search.scan()
```
TO BE TESTED!!!

ntarocco (approved):

cds_ils/migrator/documents/api.py
kprzerwa: fixed in the incoming PR

PR #164, opened 2020-10-07
global: fix config vars for deployment

ntarocco:


cds_ils/config.py
```diff
@@ -220,7 +220,7 @@ def _parse_env_bool(var_name, default=None):
 LOGGING_SENTRY_PYWARNINGS = False
 """Enable logging of Python warnings to Sentry."""
 
-LOGGING_SENTRY_CELERY = False
+LOGGING_SENTRY_CELERY = True
```
kprzerwa, line 223: 💯

kprzerwa (approved):

cds_ils/config.py
ntarocco: Actually, I will remove this from here and add it only to the deployment config

PR #168, opened 2020-10-09
override anonymous patron class

ntarocco:


tests/ldap/test_ldap_cli.py
```diff
@@ -18,18 +18,18 @@
...
-        delete_user(system_user)
+        delete_user(patron1)
         assert len(outbox) == 1
         assert outbox[0].recipients == [
             app_with_mail.config["MANAGEMENT_EMAIL"]
         ]
 
 
-def test_check_users_for_update(app, system_user):
+def test_check_users_for_update(app, patron1):
```
topless, line 32: nit: is patron1 needed here?

topless (approved):

tests/ldap/test_ldap_cli.py
ntarocco: Yes, otherwise the user is not created by the Pytest fixture.

kprzerwa (approved):

tests/ldap/test_ldap_cli.py
topless: I see its not used directly but its needed to find a user the query.

PR #169, opened 2020-10-11
eitems: serialize extra field in response

### LINKED ISSUE #162: Add proxy endpoint for e-items
At CERN, e-items can contain link that are not freely accessible, for example for ebooks that have been purchased by the library. The librarian will always mark such ebooks as `open_access: False`.

There are 2 types of closed access e-items:
1. the URL is accessible as anonymous, but then the provider/vendor will require login in their website
2. the URL requires a reverse proxy before (while will require CERN login). The link to the e-book should be prefixed withg the EzProxy URL. The URL will be marked as `login_required: True`

## Implementation

- [x] REST APIs: on `login_required: True`, when serializing back the response, a new field `login_required_url` should contain the final URL with EzProxy
- [x] UI: change every UI cmp where the link with `login_required` is shown to use the new field `login_required_url` as value instead of the original.
- [x] Adapt the migration script to set the right `open_access` and `login_required` fields for URLs.

## Example

Record metadata
```
url: {
    value: "https://myprotectedurl.com?id=...",
    login_required: True
}
```

REST API response
```
url: {
    value: "https://myprotectedurl.com?id=...", 
    login_required: True,
    login_required_url: "https://ezproxy.cern.ch?url=<encoded https://myprotectedurl.com?id=..."
}
```

Link
```
<a href="https://ezproxy.cern.ch?url=<encoded https://myprotectedurl.com?id=...">https://myprotectedurl.com?id=...</a>
```

ntarocco:
* serialize the extra field `login_required_url` when `login_required`
  is True.
* prefix the protected URL with the EzProxy URL
* partially closes #162

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -26,9 +26,12 @@ from invenio_app_ils.circulation.utils import circulation_can_be_requested, \
 from invenio_app_ils.config import \
     CELERY_BEAT_SCHEDULE as ILS_CELERY_BEAT_SCHEDULE
 from invenio_app_ils.config import RECORDS_REST_ENDPOINTS
+from invenio_app_ils.documents.api import DOCUMENT_PID_TYPE
+from invenio_app_ils.eitems.api import EITEM_PID_TYPE
 from invenio_app_ils.ill.api import can_item_circulate, \
     circulation_default_extension_duration, \
     circulation_default_loan_duration
+from invenio_app_ils.literature.api import LITERATURE_PID_TYPE
 from invenio_app_ils.locations.api import LOCATION_PID_TYPE
 from invenio_app_ils.patrons.api import PATRON_PID_TYPE, AnonymousPatron
 from invenio_app_ils.permissions import authenticated_user_permission, \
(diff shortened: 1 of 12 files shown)


cds_ils/eitems/serializers/custom_fields.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""EItem custom serializer functions."""
+
+from flask import current_app
+
+
+def field_urls(metadata):
```
kprzerwa, line 13: nit: naming police 🚔 
can we name the function with something suggesting that it actually transforms the login required urls? f.e. `format_login_required_urls` ?

cds_ils/config.py
```diff
@@ -589,3 +602,13 @@ def _parse_env_bool(var_name, default=None):
...
 
 ILS_PATRON_ANONYMOUS_CLASS = AnonymousPatron
+
+###############################################################################
+# CDS-ILS configuration
+###############################################################################
+#: LDAP configuration
+CDS_ILS_LDAP_URL = "ldap://xldap.cern.ch"
+#: Literature covers Syndetic client ID
+CDS_ILS_SYNDETIC_CLIENT = "CHANGE_ME"
```
kprzerwa, line 612: thanks for fixing it!

kprzerwa (approved):

cds_ils/eitems/serializers/custom_fields.py
ntarocco: Done!

PR #171, opened 2020-10-12
global: Adds env name

jrcastro2:


FlorianCassayre (approved):
Missing a newline at the end, but it seems to work regardless.

jrcastro2:
> Missing a newline at the end, but it seems to work regardless.

Even if we can not see it it's there :/ . (I added it)

FlorianCassayre:
My bad then :+1:

ntarocco (approved):
Thanks!!! 💯

PR #175, opened 2020-10-15
[depends on invenio-accounts] auth: redirect to CERN after logout

kprzerwa:


cds_ils/authentication/views.py
```diff
@@ -0,0 +1,18 @@
...
+
+from flask import Blueprint, current_app, redirect
+
+cern_oauth_blueprint = Blueprint("cern_openid_oauth", __name__)
+
+
+@cern_oauth_blueprint.route('/cern_openid/logout/')
+def logout_redirect():
+    """Redirect to the CERN OpenID logout."""
+    return redirect(current_app.config['OAUTH_REMOTE_REST_APP']['logout_url'])
```
ntarocco, line 18: Is this still needed? Wasn't it enough to define the `SECURITY_ ...` var with the url?

kprzerwa: it is needed necause flask security expects the urls within the application host, so redirect to keycloak is marked as invalid :)

ntarocco (approved):

cds_ils/cli.py
```diff
@@ -66,11 +66,39 @@ def pages():
...
+        "saturday",
+        "sunday",
+    ]
+
+    closed = ["saturday", "sunday"]
+    times = [
+        {"start_time": "08:00", "end_time": "12:00"},
+        {"start_time": "13:00", "end_time": "18:00"},
+    ]
+    opening_weekdays = []
```
topless, line 83: nit: I noticed an inconsistency in the tense of the verbs `closed` past, vs `opening` indicates duration etc, maybe `open` and `open_weedkdays` is more "correct"

ntarocco (approved):

topless (approved):

cds_ils/cli.py
topless: I see its comming from ILS you can ignore my comment.

PR #183, opened 2020-10-20
importer: business logic 

kprzerwa:
it is a prototype for MARCXML providers importer

logic:

receive XML ->
select one record -> 
choose/validate provider -> 
translate to JSON using rules -> 
check if create or update -> 
create

I have separated the solution from invenio-migrator, as it seems safer to use it this way. Especially because invenio-migrator assumes we migrate record to record, while we are creating many record types on cds-ils. Additionally, the main goal here is **importing**.

PROs:
1. it can recognise the provider via MARC fields
2. no need for creating new logic, we can use existing logic from cds-dojson
3. moving the rules out of the cds_dojson and adding then into the cds-ils (dojson shouldn't know about our local data model)
4. potentially easy to plug in new rules and providers
5. we can use the solution both for syncing with CDS and for importing from the other providers -  no reinventing the wheel for syncing, it will work basically the same as the migrator

CONs:
1. cds-ils grows in code as an overlay
2. a bit of learning curve(although after separating it from the invenio-migrator should be a bit easier to understand the logic)
3. a little dependency on cds-dojson

## UPDATE

rules and cleaning of the migrator dependencies to follow in a separate PR


cds_ils/importer/XMLRecordDump.py
```diff
@@ -0,0 +1,75 @@
...
+        dojson_model=marc21,
+    ):
+        """Initialize."""
+        super(self.__class__, self).__init__(
+            data, source_type, latest_only, pid_fetchers, dojson_model
+        )
+
+
+    def dump(self):
+        dt = datetime.datetime.now()
```
ntarocco, line 41:
```suggestion
        dt = datetime.datetime.utcnow()
```
UTC or server time?

cds_ils/importer/cli.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@importer.command()
+@click.argument("sources", type=click.File("r"), nargs=-1)
+@click.option(
+    "--source-type",
+    "-t",
+    type=click.Choice(["json", "marcxml"]),
+    default="marcxml",
+    help="Import from JSON, MARCXML or CDS-Migrator-Kit's _records.json file.",
```
ntarocco, line 19: this is great! Maybe someone can help and create a very simple UI to upload the file and report results.

cds_ils/importer/providers/cds/cds_ignore_fields.py
```diff
@@ -0,0 +1,95 @@
+CDS_IGNORE_FIELDS = {
```
ntarocco, line 1: this is great! 💯

cds_ils/importer/providers/cds/rules/document.py
```diff
@@ -0,0 +1,948 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: I think I will skip to check in details this, but this mean that you can remove the rules from `cds-dojson`?
Honestly I like the idea of having this here.

cds_ils/importer/providers/ebl/ebl.py
```diff
@@ -0,0 +1,10 @@
+from cds_ils.importer.base import Base
+
+
+class EBLModel(Base):
+
+    __query__ = ''
+
+    __schema__ = 'https://127.0.0.1:5000/schemas/documents/document-v1.0.0.json'
```
ntarocco, line 8:
```suggestion
    __schema__ = 'https://127.0.0.1:5000/schemas/documents/document-v1.0.0.json'
```
watch out to the HOST here, it is different when deployed. Please use the config `JSONSCHEMA_HOST`. You can use utilities if needed, see [here](https://github.com/inveniosoftware/invenio-jsonschemas/blob/master/invenio_jsonschemas/ext.py#L138)

cds_ils/importer/providers/springer/rules/document.py
```diff
@@ -0,0 +1,84 @@
+from cds_dojson.marc21.fields.books.errors import UnexpectedValue
+from cds_dojson.marc21.fields.utils import filter_list_values, clean_val, \
+    out_strip
+from dojson.utils import force_list, for_each_value
+
+from cds_ils.importer.errors import ManualImportRequired
+from cds_ils.importer.providers.springer.springer import model
+
+
+@model.over('identifiers', '^020__', )
```
ntarocco, line 10: nice! did you get this from CDS?

cds_ils/importer/XMLRecordDump.py
```diff
@@ -0,0 +1,74 @@
...
+
+        marc_record = create_record(self.data)
+
+        try:
+
+            # MARCXML -> JSON fields translation
+            val = self.dojson_model.do(
+                marc_record, exception_handlers=exception_handlers
+            )
+            missing = self.dojson_model.missing(marc_record)
```
ntarocco, line 56: can you maybe describe better, with comments or better naming what `missing` means?
Missing required fields? Missing something else?

Ok, from the log it looks like rules missing. Does it mean that there are fields in the input Marc that are taken into account in the rules? I would just be a little bit more explicit.

cds_ils/importer/api.py
```diff
@@ -0,0 +1,84 @@
...
+    except Exception as e:
+        db.session.rollback()
+        raise e
+
+
+def import_record(data, provider, source_type=None, eager=False):
+    """Import record from dump."""
+
+    source_type = source_type or 'marcxml'
+    assert source_type in ['marcxml', 'json']
```
ntarocco, line 34: does it really support `json`?

cds_ils/importer/api.py
```diff
@@ -0,0 +1,84 @@
...
+        return process_dump(data, provider, source_type=source_type)
+    else:
+        process_dump.delay(data, provider, source_type=source_type)
+
+
+def import_from_xml(sources, source_type, provider, eager=True):
+    """Load xml file."""
+
+    for idx, source in enumerate(sources, 1):
+        click.echo(
```
ntarocco, line 46: in `api.py`, you don't know that you are in CLI. I would suggest to remove these `click.echo` and replace them with a configurable logger printing to console instead (if not already the case by default in Invenio).
Same for all other files.
In the end, all this should be called via a UI form, and output displayed in a web page.

If you want to keep the colors, you might have to pass a `logger_func` in the params, which corresponds to `click.echo` when called from the CLI.

cds_ils/importer/cli.py
```diff
@@ -0,0 +1,31 @@
...
+    "-t",
+    type=click.Choice(["json", "marcxml"]),
+    default="marcxml",
+    help="Import from JSON, MARCXML",
+)
+@click.option(
+    "--provider",
+    "-p",
+    type=click.Choice(["springer", "cds", "ebl"]),
+    help="Import from JSON, MARCXML",
```
ntarocco, line 26:
```suggestion
    help="Set the name of the provider",
```

cds_ils/importer/config.py
```diff
@@ -0,0 +1,13 @@
+CDS_ILS_IMPORTER_RECORD_TAG = "//*[local-name() = 'record']"
+
+CDS_ILS_IMPORTER_PROVIDERS = {
+    'cds': {'priority': 1,
+            'agency_code': 'SZeGeCERN',
```
ntarocco, line 5: is `agency_code` something defined in MarcXML?

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+
+from cds_ils.importer.errors import DocumentImportError
+
+
+def check_search_results(result, search_term, search_term_name,
+                         raise_on_empty=False, raise_on_many=True):
+    """Check number of search results"""
+    hits_total = result.hits.total.value
+
+    if hits_total < 1:
```
ntarocco, line 15:
```suggestion
    if hits_total == 0:
```

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+
+        click.secho(
+            "no document found with {} {}".format(search_term_name,
```
ntarocco, line 18:
```suggestion
            "no document found with `{}:{}`".format(search_term_name,
```

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+            raise DocumentImportError(
+                "found more than one document with {} {}".format(
+                    search_term_name, search_term)
+            )
+    return hits_total
+
+
+def search_documents_by_isbn(isbn):
+    """Find document by ISBN."""
+    search = DocumentSearch().query(
```
ntarocco, line 43:
```suggestion
    DocumentSearch = current_app_ils.document_search_cls
    search = DocumentSearch().query(
```
And for all of them around.

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+    return search.execute()
+
+
+def get_document_by_exact_title_authors(title, authors):
+    """Find document by title and authors."""
+
+    search = DocumentSearch().query(
+        "bool",
+        must=[
+            Q("match", title=title),
```
ntarocco, line 71: I think that for **exact** match, you need to use `term` queries:
```suggestion
            Q("term", title=title),
```
but I might be wrong, to be checked (see [here](https://stackoverflow.com/questions/40124317/what-is-the-difference-between-a-term-query-and-a-match-one). With `match`, you might get results ignoring upper/lower cases, but also on other cases because the match will be done on the analyzed terms.
Or maybe you want this, but then the function name should be different...
```

File: cds_ils/importer/documents/api.py [hunk 16]
```diff
@@ -0,0 +1,128 @@
...
+            )
+    return hits_total
+
+
+def search_documents_by_isbn(isbn):
+    """Find document by ISBN."""
+    search = DocumentSearch().query(
+        "bool",
+        must=[
+            Q("match", identifiers__scheme="ISBN"),
```
ntarocco, line 46: For all of them:
```suggestion
            Q("term", identifiers__scheme="ISBN"),
```
See my comment below.

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+
+    search = DocumentSearch().query(
+        "bool",
+        must=[
+            Q("match", title=title),
+            Q("terms", authors__full_name=authors),
+        ],
+    )
+
+    result = search.execute()
```
ntarocco, line 76: I would remove the `if` just after, at least it looks like that you expect a return obj each time.
```suggestion
return search.execute()
```

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def fuzzy_search_document(title, authors):
+    build_fuzzy_authors_query = [Q({"fuzzy":
+        {
+            "authors.full_name": {
+                "value": author,
+                "fuzziness": "AUTO",
+                "prefix_length": 0,
+                "transpositions": True,
```
ntarocco, line 88: what about adding a very small comment explaining what each of this thing does, for newbies like me? :) :D

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,128 @@
...
+            }),
+            Bool(
+                "bool",
+                should=build_fuzzy_authors_query
+            )
+        ],
+    )
+    result = search.execute()
+    if len(result) > 0:
+        return [x.pid for x in result.hits]
```
ntarocco, line 116: you are going to get only the first 10/15...
In general, I would suggest to always return `search.execute()`, or even only `search` so you can choose to use `execute()` or `scan()` for example when used.
These are APIs methods, they should be slightly more generic without knowing who is calling them.

cds_ils/importer/documents/importer.py
```diff
@@ -0,0 +1,110 @@
...
+        return cleaned
+
+    def create_document(self):
+        """Create a new record from dump."""
+        cleaned_json = self._before_create()
+        document_class = current_app_ils.document_record_cls
+        # Reserve record identifier, create record and recid pid in one
+        # operation.
+        record_uuid = uuid.uuid4()
+        provider = DocumentIdProvider.create(
```
ntarocco, line 43: When syncing with CDS (not migration), we might what to store somewhere the CDS recid (which is not the `legacy_id`) to show a button `See this in CDS` or something, WDYT?
Actually, it might be already done in `alternative_identifiers`? Would be great to have a test for this.

cds_ils/importer/documents/importer.py
```diff
@@ -0,0 +1,110 @@
...
+            object_uuid=record_uuid,
+        )
+        cleaned_json["pid"] = provider.pid.pid_value
+
+        try:
+            document = document_class.create(cleaned_json, record_uuid)
+            db.session.commit()
+            return document
+        except IlsValidationError as e:
+            click.secho("Field: {}".format(e.errors[0].res["field"]), fg="red")
```
ntarocco, line 54: do we have to `rollback` here too? Otherwise the PID is almost minted, but not the doc... or maybe:
```
with db.session.nested:
  PID
  Record
db.session.commit
or 
rolllback
```

cds_ils/importer/documents/importer.py
```diff
@@ -0,0 +1,110 @@
...
+                    identifier['scheme'] == 'DOI']
+
+        authors = [author['full_name'] for author in
+                   self.json_data.get('authors', [])]
+
+        matches = []
+
+        # check by isbn first
+        for isbn in isbn_list:
+            results = search_documents_by_isbn(isbn)
```
ntarocco, line 95: first 10/15 only, is it ok?
Same below...

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,116 @@
...
+            existing_eitem.update(metadata_to_update)
+            existing_eitem.commit()
+        else:
+            for hit in results:
+                existing_eitem = EItem.get_record_by_pid(hit.pid)
+                existing_eitem.delete()
+            self.import_eitem(matched_document)
+
+        if self.delete_lower_priority_providers:
+            document_eitems = EItemSearch().search_by_document_pid(
```
ntarocco, line 72: It is probably preferable to search in the PID Store db tables when using the PID directly, not ES.
Please double check this in the entire code.

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def update_eitems(self, matched_document):
+        """Update eitems for given document."""
+        import ipdb; ipdb.set_trace()
+        document_pid = matched_document["pid"]
+
+        results = get_eitems_for_document_by_provider(document_pid,
+                                                      self.metadata_provider)
+
+        if len(results) == 0:
```
ntarocco, line 56: We should always use `search.hits`. Please do not use `len` because you are going to get only the first 10/15.
In this case it is ok checking for `0`, but better to have good examples in the code.
We should check this everywhere...

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,116 @@
...
+                doi['scheme'] == 'DOI']
+        eitem_json.update({"document_pid": document_pid,
+                           "open_access": self.open_access,
+                           "identifiers": dois,
+                           "created_by": {"type": "import",
+                                          "value": self.metadata_provider}
+                           })
+
+    def _apply_url_login(self, eitem):
+        if 'urls' in eitem:
```
ntarocco, line 94:
```suggestion
        for url in eitems.get('urls', []):
```

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,116 @@
...
+        if eitem_json:
+            self._build_eitem_dict(eitem_json, new_document["pid"])
+            record_uuid = uuid.uuid4()
+            provider = EItemIdProvider.create(
+                object_type="rec",
+                object_uuid=record_uuid,
+            )
+
+            eitem_json["pid"] = provider.pid.pid_value
+            eitem = EItem.create(eitem_json, record_uuid)
```
ntarocco, line 112: can it fail validation? Better to rollback just in case? See my comment above.

cds_ils/importer/api.py
kprzerwa: not yet, I will remove.

cds_ils/importer/api.py
kprzerwa: I left it for later, there is a separate task for logging in the importer. I had in mind to do it similarly as the reports for the user sync. WDYT?

cds_ils/importer/XMLRecordDump.py
kprzerwa: this naming is coming from cds-dojson, can't refactor it now since cds-videos is using it too, but I will leave a comment here

cds_ils/importer/config.py
kprzerwa: yes it is an id of the agency keeping the repository, every XML coming from the providers has its own code

cds_ils/importer/providers/ebl/ebl.py
kprzerwa: the json schema host is solved by dojson, I just didn't clean this part yet

cds_ils/importer/documents/api.py
kprzerwa: I've tested them and changed accordingly

cds_ils/importer/documents/api.py
kprzerwa: 👍

cds_ils/importer/documents/importer.py
kprzerwa: I thought it would be the `legacy_recid`. Did you have something else in mind?

cds_ils/importer/eitems/importer.py
kprzerwa: for the EItems we are not minting the `document_pid` as PID, I'm not sure if I understand your comment

cds_ils/importer/eitems/importer.py
kprzerwa: added the try/except block

cds_ils/importer/documents/importer.py
kprzerwa: fixed all occurences

cds_ils/demo_data/items.json
```diff
@@ -1,6 +1,10 @@
 [
   {
     "pid": "678e3-an678A",
+    "created_by": {
```
ntarocco, line 4: This is also in the other PR of yours, might have conflict

cds_ils/importer/XMLRecordDump.py
```diff
@@ -0,0 +1,83 @@
...
+from flask import current_app
+from invenio_pidstore.resolver import Resolver
+from invenio_records.api import Record
+
+from cds_ils.importer import marc21
+from cds_ils.importer.errors import LossyConversion
+from cds_ils.importer.handlers import importer_exception_handler
+
+
+class XMLRecordDump(object):
```
ntarocco, line 24: Just IMHO: I would have called this more `XMLRecordToJson` :)

cds_ils/importer/api.py
```diff
@@ -0,0 +1,91 @@
...
+                        report["updated"],
+                        report["ambiguous"],
+                        report["fuzzy"],
+                    ),
+                    fg="blue",
+                )
+
+        except IlsValidationError as e:
+            records_logger.error(
+                "@FILE: {0} FATAL: {1}".format(
```
ntarocco, line 82: Just to keep it in mind: when deployed, there is no log file.
If it makes sense to keep track of these logs, better to have a JSON format, so we can collect them and send to central logging, as done in ILS (see the indexer class)

cds_ils/importer/config.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Copyright (C) 2020 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""CDS-ILS Importer module."""
+CDS_ILS_IMPORTER_RECORD_TAG = "//*[local-name() = 'record']"
+
+CDS_ILS_IMPORTER_PROVIDERS = {
```
ntarocco, line 11: what about to put in the main `config.py` file, so that you can use it every where with `current_app.config`?
If it is not a config (overridable at some point), then it is more constants? I would change the file name.

cds_ils/importer/documents/api.py
```diff
@@ -0,0 +1,130 @@
...
+
+def get_document_by_legacy_recid(legacy_recid):
+    """Search documents by its legacy recid."""
+    document_search = current_app_ils.document_search_cls()
+    document_cls = current_app_ils.document_record_cls
+
+    search = document_search.query(
+        "bool", filter=[Q("term", legacy_recid=legacy_recid)]
+    )
+    result = search.execute()
```
ntarocco, line 126: Maybe count is enough?
```suggestion
    result = search.count()
```
it should simply give you the number back.

cds_ils/importer/documents/importer.py
ntarocco: Probably more `alternative_identifiers` instead of adding a new field. We might need to update the vocabulary.

cds_ils/importer/documents/importer.py
ntarocco: maybe you have missed this comment?

cds_ils/importer/eitems/api.py
```diff
@@ -0,0 +1,19 @@
...
+from invenio_app_ils.proxies import current_app_ils
+
+
+def get_eitems_for_document_by_provider(document_pid, provider):
+    """Find eitem by document pid and provider."""
+    eitem_search = current_app_ils.eitem_search_cls()
+    search = eitem_search.search_by_document_pid(
+        document_pid=document_pid
+    ).filter("term", created_by__value=provider)
+    result = search.execute()
```
ntarocco, line 18: execute? or scan/count? You get only the first 15.

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,169 @@
...
+        elif hits_total == 1:
+            try:
+                existing_eitem = eitem_cls.get_record_by_pid(
+                    results.hits[0].pid)
+                metadata_to_update = {}
+                self._build_eitem_dict(metadata_to_update,
+                                       matched_document["pid"])
+                existing_eitem.update(metadata_to_update)
+                existing_eitem.commit()
+                eitem_indexer.index(existing_eitem)
```
ntarocco, line 93: indexing before `db.session.commit` might create the issue that if the db commit fails, it is already indexed. What about collecting all the records to index in a list and index at the end after the db commit?

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,169 @@
...
+            except IlsValidationError as e:
+                db.session.rollback()
+                click.secho("Field: {}".format(e.errors[0].res["field"]),
+                            fg="red")
+                click.secho(e.original_exception.message, fg="red")
+        else:
+            for hit in results:
```
ntarocco, line 100: it is not clear to me why we would delete in this case...
Can you refactor with a named method or add a comment?

cds_ils/importer/eitems/importer.py
```diff
@@ -0,0 +1,169 @@
...
+                existing_eitem = eitem_cls.get_record_by_pid(hit.pid)
+                existing_eitem.delete()
+            self.import_eitem(matched_document)
+
+        if self.delete_lower_priority_providers:
```
ntarocco, line 105: ok I am totally lost now :)
Sorry, probably just me... No need to explain here in the comment, but I think the code should be maybe a little bit more self explanatory?
Maybe it require a little bit more methods/objects that explain behaviour? Why/in which case delete or import?

ntarocco (commented):
If you prefer, maybe you can merge this PR and address the comments that you consider valid in the next one?

cds_ils/demo_data/items.json
kprzerwa: it was in invenio-app-ils

cds_ils/importer/api.py
kprzerwa: yes, sure, there is a separate task to clean the logging of the importer

cds_ils/importer/documents/api.py
kprzerwa: but I would like to have the actual result, not the count only.
Anyway this will be changed after Konstantina's PR with the legacy_recid minter

cds_ils/importer/eitems/importer.py
kprzerwa: sure I will refactor this part

cds_ils/importer/eitems/importer.py
kprzerwa: we can quickly chat about it, I have doubts if it is a good solution now ;)

cds_ils/importer/documents/importer.py
kprzerwa: but we already have the `legacy_recid` field for the migration, I'm not sure if I understand what do you mean

Issue #80, opened 2019-09-27
CERN configuration to override ILS


ntarocco:
Double check the full `config.py` in `invenio-app-ils` and `invenio-circulation`.
To customize:
* all `CIRCULATION_POLICIES`
    * [x] `circulation_default_loan_duration`: items have different circulation restrictions and some can loaned only for 2 weeks, some for 4 weeks. This should check the `item` restriction and return the right duration. Same for `duration_validate`
  * [ ] ILS_LOAN_WILL_EXPIRE_DAYS
* E-mail config and templates
  * [ ] ILS_MAIL_ENABLE_TEST_RECIPIENTS
  * [ ] ILS_MAIL_NOTIFY_TEST_RECIPIENTS
  * [x] ILS_MAIL_LOAN_TEMPLATES (No longer exists)
  * [x] ILS_MAIL_DOCUMENT_REQUEST_TEMPLATES (No longer exists)
  * [x] ILS_MAIL_LOAN_MSG_LOADER (No longer exists)
  * [x] ILS_MAIL_LOAN_OVERDUE_REMINDER_INTERVAL (No longer exists)
  * [x] all email templates ([document](https://cernbox.cern.ch/index.php/apps/files/?dir=/__myshares/CDS%20-%20Library%20shared%20(id:294118))), in particular fix [list_loans](https://github.com/inveniosoftware/invenio-app-ils/blob/45b844720f65af0533f1bf7725cd9154ea5a3973/invenio_app_ils/circulation/tasks.py#L58) one (ask Nico :))
* [x] Locations: block creation and deletions
* [x] Circulation transitions: remove "In transit"
* `invenioConfig.js`
  * [ ] extensionsMaxCount
  * [ ] loanWillExpireDays
  * [ ] ... any other that makes sense ... ask the library!
* [ ] delivery methods

jrcastro2:
- Email config checked 
- [x] MAIL_NOTIFY_BCC
- [x] MAIL_NOTIFY_CC
- [x] ILS_CIRCULATION_MAIL_OVERDUE_REMINDER_INTERVAL
- [x] SECURITY_CONFIRM_EMAIL_WITHIN
- [x] SECURITY_SEND_REGISTER_EMAIL
- [x] SECURITY_SEND_PASSWORD_CHANGE_EMAIL
- [x] SECURITY_SEND_PASSWORD_RESET_EMAIL
[...]

PR #185, opened 2020-10-20
email: Adds email templates

jrcastro2:
partially closes https://github.com/CERNDocumentServer/cds-ils/issues/80


cds_ils/templates/invenio_app_ils_circulation/mail/cds_active_loans.html
```diff
@@ -0,0 +1,28 @@
+{% block subject %}
+InvenioILS: Active loans of patron "{{ patron.name }}" with pid "{{ patron.pid }}".
+{% endblock %}
+
+{% block body_plain %}
+Dear librarians,
+
+when removing users who are no longer registered in the system, it has been found a patron with active loans.
```
KonstantinaStoikou, line 8: Maybe something like ` a patron has been found with active loans` sounds better?

KonstantinaStoikou (approved):

cds_ils/templates/invenio_app_ils_circulation/mail/cds_active_loans.html
kprzerwa: it will be approved or rewritten by the library so I suggest its better not to change content for now

kprzerwa (approved):

ntarocco (approved):

cds_ils/config.py
```diff
@@ -432,7 +432,36 @@ def _parse_env_bool(var_name, default=None):
 }
 
 ILS_CIRCULATION_MAIL_TEMPLATES = {
-    "librarian_footer": "cds_librarian_footer.html"
+    "librarian_footer": "cds_librarian_footer.html",
```
ntarocco, line 435: I think this should disappear no? We can merge it and you remove it later on if you prefer...

jrcastro2: I kept it for now because it requires some refactor in `invenio-app-ils` if we want to remove it from there and add the footer to the `active_loans.html` template. IMHO it's better to remove it later

PR #186, opened 2020-10-21
StaticPages: add content for static pages

KonstantinaStoikou:
Selection_036
Selection_037

### DIFF
--- a/cds_ils/cli.py
+++ b/cds_ils/cli.py
@@ -7,8 +7,10 @@
 
 """CLI for CDS-ILS."""
 import os
+import pathlib
 
 import click
+import pkg_resources
 from flask import current_app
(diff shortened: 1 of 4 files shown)


cds_ils/cli.py
```diff
@@ -31,26 +32,35 @@ def fixtures():
 @with_appcontext
 def pages():
     """Register CDS static pages."""
+    CURRENT_DIR = pathlib.Path(__file__).parent.absolute()
```
kprzerwa, line 35: there is slightly more "elegant" way to do it
https://github.com/CERNDocumentServer/cds-videos/blob/cdslabs_qa/cds/modules/fixtures/cli.py#L228f

kprzerwa (approved):

PR #193, opened 2020-10-22
ldap: fix + refactor sync users

ntarocco:
* refactors LDAP sync and import to avoid db session close after user
  deletion
* change APP_ALLOWED_HOSTS config to be more secure
* change cron task to run overnight

### DIFF
--- a/cds_ils/ldap/api.py
+++ b/cds_ils/ldap/api.py
@@ -9,6 +9,7 @@
 """CDS-ILS ldap API."""
 
 import json
+import time
 import uuid
 
 import ldap
@@ -19,9 +20,13 @@ from invenio_app_ils.circulation.tasks import send_active_loans_mail
 from invenio_db import db
 from invenio_oauthclient.models import RemoteAccount, UserIdentity
 from invenio_userprofiles.models import UserProfile
-from sqlalchemy.orm.exc import NoResultFound
 
 from cds_ils.config import OAUTH_REMOTE_APP_NAME
+
+
+def ldap_user_get(user, field_name):
+    """Get first value of the given field from the LDAP user object."""
+    return user[field_name][0].decode("utf8")
 
 
 class LdapClient(object):
@@ -39,6 +44,10 @@ class LdapClient(object):
         ]
     """
 
+    LDAP_BASE = "OU=Users,OU=Organic Units,DC=cern,DC=ch"
+
+    LDAP_CERN_PRIMARY_ACCOUNTS_FILTER = "(&(cernAccountType=Primary))"
+
     LDAP_USER_RESP_FIELDS = [
         "mail",
         "displayName",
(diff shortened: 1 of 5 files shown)


cds_ils/ldap/api.py
```diff
@@ -146,196 +135,254 @@ class LdapUserImporter:
...
         """Constructor."""
-        self.ldap_users = ldap_users
         self.client_id = current_app.config["CERN_APP_OPENID_CREDENTIALS"][
             "consumer_key"
         ]
 
-    def import_user_identity(self, user_id, ldap_user):
+    def create_invenio_user(self, ldap_user):
+        """Commit new user in db."""
+        email = ldap_user_get(ldap_user, "mail").lower()
```
kprzerwa, line 146: 🚀

cds_ils/ldap/api.py
```diff
@@ -146,196 +135,254 @@ class LdapUserImporter:
...
+        remote_account = self.create_invenio_remote_account(user_id, ldap_user)
+        db.session.add(remote_account)
 
 
-def import_ldap_users(ldap_users):
+def _reindex_patrons():
+    """Launch the CLI cmd to re-index all users."""
+    from invenio_base.app import create_cli
+
+    cli = create_cli()
```
kprzerwa, line 196: not necessarily for this PR, because we use it everywhere but this does not log exceptions (I tried with `catch_exceptions`, and I think it is not working - for me indexing invoked with this was failing silently)

cds_ils/ldap/api.py
```diff
@@ -146,196 +135,254 @@ class LdapUserImporter:
...
 
 
-def delete_user(user_id):
-    """Delete a system user."""
-    with current_app.app_context():
-        try:
-            anonymize_patron_data(user_id)
-        except AssertionError:
-            send_active_loans_mail(user_id)
+def _delete_invenio_user(user_id):
```
kprzerwa, line 239: do we delete the RemoteAccounts and UserProfile too ?

kprzerwa (approved):

cds_ils/ldap/api.py
ntarocco: ok interesting, I trusted that code but it is true that I did not verify it. I need to check it because it is important to re-index.

cds_ils/ldap/api.py
ntarocco: not sure I understand the comment: yes, `anonymize_patron_data` will delete everything, all the user related rows in the various db tables.

cds_ils/ldap/api.py
```diff
@@ -146,196 +135,254 @@ class LdapUserImporter:
...
+    """Delete an Invenio user."""
+    try:
+        anonymize_patron_data(user_id)
+        return True
+    except AssertionError:
+        send_active_loans_mail(user_id)
```
ntarocco, line 245: @jrcastro2 you will have to refactor this line, just be aware of the changes, start from the PR and not from master

cds_ils/ldap/api.py
ntarocco: as agreed, I will fix in the next PR, it needs changed in `invenio-app-ils` too

PR #205, opened 2020-10-28
document: Updates extra fields

jrcastro2:
Updates the extra fields of the document to add the line attribute and to remove unnecessary lists.
Related to https://github.com/inveniosoftware/react-invenio-app-ils/pull/253


cds_ils/config.py
```diff
@@ -417,15 +417,15 @@ def _parse_env_bool(var_name, default=None):
...
+            "marshmallow": SanitizedUnicode(),
         },
         "unit:project": {
             "elasticsearch": "keyword",
             "marshmallow": SanitizedUnicode(),
         },
         "unit:study": {
             "elasticsearch": "keyword",
-            "marshmallow": List(SanitizedUnicode()),
+            "marshmallow": SanitizedUnicode(),
```
topless, line 428: @kprzerwa do you bless this change?

topless (approved):

cds_ils/config.py
jrcastro2: This change was motivated by the answer from the librarians `Library:There is no KB in CDS yet.` can be checked in [CodiMD](https://codimd.web.cern.ch/uR0k-4yiTzW0BSnp7mDw0Q#Institutions)

kprzerwa: Sorry I don't follow, they said that there is no KB yet, so you change here from List to string?
I'm afraid I do not bless this change (neither the previous one, with the institutions), @ntarocco do you know anything more about this that I am not aware of?

IMO it should be a list, what if the record is connected with more than one institution or study ? It can happen, like in case of experiments and accelerators

ui/src/config.js
```diff
@@ -71,59 +71,70 @@ export const config = {
           type: 'vocabulary',
           vocabularyType: 'document_accelerators',
           default: '',
-        },
-        'unit:curated_relation': {
-          label: 'Curated Relation',
-          type: 'boolean',
-          default: false,
+          line: 1,
```
kprzerwa, line 74: now I understood the sorting from the previous PR. You've defined the sorting function in `react-invenio-app-ils`, do we have there any extra fields (as a show case)? I'm asking because I didn't see the changed configuration there

jrcastro2: There are no changes in `react-invenio-app-ils` because the extra fields are [empty](https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/config/defaultConfig.js#L224)

ui/src/config.js
```diff
@@ -71,59 +71,70 @@ export const config = {
...
         },
         'unit:experiment': {
           label: 'Experiment',
           type: 'vocabulary',
           vocabularyType: 'document_experiments',
           default: '',
+          line: 1,
```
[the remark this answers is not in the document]
jrcastro2: Hmm, I understand your point but if we add another component that wants to display them differently, it can just omit the `line` attribute. When I added this `line` attribue, I had in mind to make it easy to any user of invenio to organize the extra fields easily when creating their app. As I understand it, we have this configuration object because for each invenio based project we will have different extra fields, so we should be able to organize the fields acordingly to the project. In our case, in `cds` it makes sense to have 3 fields in the first line but maybe others want only 1 field per line. This was the motivation behind the change, it doesn't feel right to me to provide a static layout for dynamic fields.

kprzerwa (approved):

ui/src/config.js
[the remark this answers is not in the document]
ntarocco: Agree with your comments, but it does not feel right either to put presentation in configuration: if you put an "order" key, then it means that an order is important when presenting, but it is not defining how you present it.
If you put `line`, you are really meaning that these will be show in lines, and not cols, buttons, labels, etc...

PR #206, opened 2020-10-28
release: v1.0.0a18

kprzerwa:


docker-services.yml
```diff
@@ -35,6 +35,10 @@ services:
       - bootstrap.memory_lock=true
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
       - discovery.type=single-node
+      - cluster.routing.allocation.disk.threshold_enabled=true
```
kprzerwa, line 38: I've added this since local ES was failing all the time

ntarocco: Something to be added to `invenio-app-ils` too? And in [docker-services-cli](https://github.com/inveniosoftware/docker-services-cli/blob/master/docker_services_cli/docker-services.yml#L38)?

ntarocco (approved):

docker-services.yml
kprzerwa: I'm not sure if the others have the same issues already. Maybe should be tested first by more people ?

Issue #182, opened 2020-10-20
Set up resources for the application services


kprzerwa:
Calculate and assign correct resources (CPU/RAM) for the services.
Set up the number DB connections

- rabbit-mq: at least 1024 MB memory
- celery worker: 512 MB is not enough

kprzerwa:
closing as a duplicate

PR #208, opened 2020-10-29
[DO NOT MERGE] debug ldap

FlorianCassayre:
* add ipdb as requirement
* additional logs in ldap update

### DIFF
--- a/cds_ils/ldap/api.py
+++ b/cds_ils/ldap/api.py
@@ -306,16 +306,9 @@ def update_users():
         "users_cached",
     )
 
-    remote_accounts = RemoteAccount.query.all()
-    _log_info(
-        log_uuid,
-        "users_fetched_from_invenio",
-        dict(users_fetched=len(remote_accounts)),
-    )
-
     # get all Invenio remote accounts and prepare a list with needed info
     invenio_users = []
-    for remote_account in remote_accounts:
+    for remote_account in RemoteAccount.query:
         invenio_users.append(
             dict(
                 remote_account_id=remote_account.id,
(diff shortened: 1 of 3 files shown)


ntarocco:
Here the latest Sentry exception https://cds-sentry.web.cern.ch/sentry/cds-ils/issues/26568/, can you have a look when working on this?

FlorianCassayre:
This comes from new code of this PR. I'm trying to understand what's wrong.

FlorianCassayre:
[Here](https://gist.github.com/FlorianCassayre/f9eb4f3b06233bfebbb64739ac9ed43f) is the complete log of the error, from the worker's perspective.
Observations:
- the function `update_users` does work if executed from the debug pod, but not if executed as a task on the celery worker pod
- the log allows us to indisputably conclude that `RemoteAccount.query.all()` is the direct or indirect cause of the crash
- this query loads ~38,000 rows in memory; each row holds ~100 bytes of data. Normally this should not have been an issue
- after profiling `RemoteAccount.query.all()` (inside an invenio shell on the debug pod) I observed an increase of 131MiB (from 233MiB to 364MiB). When the returned value is garbage collected, the memory usage drops by 111MiB. Thus the memory footprint is about ~120MiB
- the celery worker pod is limited to 512MiB of memory and uses 350-400MiB on average

Conclusion: the ldap task fails because the celery worker pod runs out of memory, in loading of all remote accounts uses a surprisingly high amount of memory (but not too concerning either).

The solution to this problem is to either increase the memory limit of that pod, or to change the code to avoid fetching all users at once by using a database iterator instead. I will definitely go for the second solution, but an increase of the memory is not to be excluded (especially if we have other heavy tasks in the future).

FlorianCassayre:
Scaling the memory up fixed the problem. #182 will eventually decide what precise amount of resources is needed.

PR #210, opened 2020-10-30
ldap: update email when changed

FlorianCassayre:
* updates the email of a user if it was changed
* safeguard in case of missing email
* tests

**Note**: the issue of the task being killed was described and fixed in #208

### DIFF
--- a/cds_ils/ldap/api.py
+++ b/cds_ils/ldap/api.py
@@ -9,6 +9,7 @@
 """CDS-ILS ldap API."""
 
 import json
+import sys
 import time
 import uuid
 
@@ -30,6 +31,11 @@ from cds_ils.mail.tasks import send_warning_mail_patron_has_active_loans
 def ldap_user_get(user, field_name):
     """Get first value of the given field from the LDAP user object."""
     return user[field_name][0].decode("utf8")
+
+
+def ldap_user_get_email(user):
+    """Get the normalized email attribute from the LDAP user object."""
+    return ldap_user_get(user, "mail").lower()
 
 
 class LdapClient(object):
@@ -146,7 +152,7 @@ class LdapUserImporter:
 
     def create_invenio_user(self, ldap_user):
         """Commit new user in db."""
-        email = ldap_user_get(ldap_user, "mail").lower()
+        email = ldap_user_get_email(ldap_user)
         user = User(email=email, active=True)
         db.session.add(user)
         db.session.commit()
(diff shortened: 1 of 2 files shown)


cds_ils/ldap/api.py
```diff
@@ -279,11 +289,26 @@ def update_users():
...
-        ldap_users_map[ldap_person_id] = ldap_user
+        if "mail" not in ldap_user:
+            _log_info(
+                log_uuid,
+                "missing_email",
+                dict(employee_id=ldap_user_get(ldap_user, "employeeID")),
+            )
+            continue
+        email = ldap_user_get_email(ldap_user)
+        if email not in ldap_users_emails:
```
KonstantinaStoikou, line 303: I don't understand the purpose of the `ldap_users_emails`. Is there any chance two ldap users to have the same email?

kprzerwa: there is a chance that ldap user does not have an email

cds_ils/ldap/api.py
```diff
@@ -206,7 +211,11 @@ def import_users():
     imported = 0
     importer = LdapUserImporter()
     for ldap_user in ldap_users:
-        email = ldap_user_get(ldap_user, "mail").lower()
+        if "mail" not in ldap_user:
+            print("User with employee ID {} does not have an email"
```
kprzerwa, line 215: maybe not print but log ?

cds_ils/ldap/api.py
FlorianCassayre: Good question; unfortunately yes there are a few ldap users with the same email (ie. [<bureau@stauffersa.ch>]()), and this is what triggers an error later on. These accounts are a result of a bad management of the users (they were manually created from what I heard) and can be safely ignored. We also observed that the email could be occasionally absent from the ldap user data, which is very unfortunate. There is no explanation to that and we agreed to ignore users with no email since these would eventually have one the next time `update_users` is called.

tests/ldap/test_ldap_cli.py
```diff
@@ -182,6 +198,7 @@ def check_existence(expected_email, expected_name, expected_department):
         "ldap.user333@cern.ch", "Nothing changed", "Same department"
     )
     check_existence("ldap.user444@cern.ch", "old user left CERN", "Department")
+    check_existence("ldap.user555@cern.ch", "Name 1", "Department 1")
```
kprzerwa, line 201: can you add a test for the one without the email?

FlorianCassayre:
@kprzerwa [Here](https://cds-sentry.web.cern.ch/sentry/cds-ils/issues/26568/?environment=dev) is an occurrence of a missing email.
image

cds_ils/ldap/api.py
FlorianCassayre: I think this is supposed to be run from the command line; but I agree for the one in `import_users`.

ntarocco (approved):

FlorianCassayre:
[Travis build](https://travis-ci.com/github/FlorianCassayre/cds-ils/builds/198448566)

kprzerwa (approved):

PR #220, opened 2020-11-05
cli: recreate covers

kprzerwa:


cds_ils/cli.py
```diff
@@ -223,6 +237,44 @@ def run_command(command, catch_exceptions=False, verbose=True):
...
+def recreate_cover(pid):
+    # search doc
+    record = Document.get_record_by_pid(pid)
+    indexer = DocumentIndexer()
+    if not record:
+        record = Series.get_record_by_pid(pid)
+        indexer = SeriesIndexer()
+    if record:
+        pick_identifier_with_cover_task(record)
+        indexer.index(record)
```
ntarocco, line 249: no need to re-index here

cds_ils/cli.py
```diff
@@ -223,6 +237,44 @@ def run_command(command, catch_exceptions=False, verbose=True):
...
+@click.argument("pid", nargs=1, type=str)
+@with_appcontext
+def recreate(pid):
+    recreate_cover(pid)
+
+
+@covers.command()
+@with_appcontext
+def recreate_all_documents():
+    all_documents = DocumentSearch().scan()
```
ntarocco, line 265:
```suggestion
    all_documents = current_app_ils.document_search_cls().scan()
```
same for series, to avoid hardcoding

ntarocco: actually, would it be "safer" if we query the DB?
Get from PID store all the `docids` that are registered, in batches, and iterate. WDYT?

cds_ils/cli.py
```diff
@@ -42,11 +50,17 @@
...
 
+from cds_ils.literature.tasks import pick_identifier_with_cover_task
+
 
 @click.group()
 def fixtures():
     """Create initial data and demo records."""
 
+@click.group()
+def covers():
```
ntarocco, line 61: Just a suggestion in the CLI commands:
`invenio covers recreate docid 123`
`invenio covers recreate docid --all`
so that you have the PID type and PID value, and you don't have to "guess" it via different commands

cds_ils/cli.py
```diff
@@ -223,6 +231,38 @@ def run_command(command, catch_exceptions=False, verbose=True):
...
+@click.option('-t', '--pid-type', required=True,
+              type=click.Choice(['docid', 'serid']))
+@click.option('-a', '--all', required=False, is_flag=True)
+@click.option('-p', '--pid')
+@with_appcontext
+def recreate(pid, pid_type, all):
+    """Recreate covers command."""
+    if pid_type == 'docid':
+        record_class, search_class, indexer_class = \
+            Document, DocumentSearch, DocumentIndexer
```
ntarocco, line 254: I would use the ones configured in the config instead of hardcoding, whenever possible:
```suggestion
            current_app_ils.document_record_cls,....
```

cds_ils/cli.py
ntarocco: You might have missed my comment maybe... WDYT?

ntarocco (commented):
See my other comment on pid store table instead of ES.

cds_ils/cli.py
kprzerwa: fixed below in the command

cds_ils/cli.py
```diff
@@ -223,6 +227,43 @@ def run_command(command, catch_exceptions=False, verbose=True):
...
+    if pid_type == DOCUMENT_PID_TYPE:
+        record_class = current_app_ils.document_record_cls
+    elif pid_type == SERIES_PID_TYPE:
+        record_class = current_app_ils.series_record_cls
+    if all:
+        pid_list = (
+            PersistentIdentifier.query.filter_by(
+                object_type="rec", status=PIDStatus.REGISTERED
+            )
+            .filter(PersistentIdentifier.pid_type.in_([pid_type]))
```
ntarocco, line 258: Just a comment, no need to change, just wondering why this `list in` instead of a `==`... it might be more performant too.

ntarocco (approved):

cds_ils/cli.py
kprzerwa: that's a leftover, I had a list of pid types there before as a an argument. I will change it to equals

PR #224, opened 2020-11-06
Cli: user testing fail if item has active loan

KonstantinaStoikou:
In `user-testing prepare` command, display error and fail when trying to create the ongoing loan if someone else has an active loan on the item (but the past loan will be created).

### DIFF
--- a/cds_ils/cli.py
+++ b/cds_ils/cli.py
@@ -10,23 +10,21 @@ import json
 import os
 import pathlib
 import random
-from datetime import date, timedelta
+from datetime import timedelta
 from random import randint
 
 import arrow
 import click
-import lorem
 import pkg_resources
+from elasticsearch import VERSION as ES_VERSION
 from elasticsearch_dsl import Q
 from flask import current_app
 from flask.cli import with_appcontext
 from invenio_accounts.models import User
-from invenio_app_ils.circulation.api import \
-    circulation_default_loan_duration_for_item
+from invenio_app_ils.circulation.search import get_active_loan_by_item_pid
 from invenio_app_ils.cli import minter
 from invenio_app_ils.documents.api import DOCUMENT_PID_TYPE
 from invenio_app_ils.indexer import wait_es_refresh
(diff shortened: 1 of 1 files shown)


cds_ils/cli.py
```diff
@@ -294,69 +292,94 @@ def create_demo_items(items_path):
...
+            "patron_pid": "{}".format(patron_pid),
+            "pickup_location_pid": "{}".format(loc_pid),
+            "transaction_location_pid": "{}".format(loc_pid),
+            "transaction_user_pid": "{}".format(patron_pid),
+            "delivery": {"method": delivery},
+        }
+
+        if is_past_loan:
+            loan_dict["state"] = "ITEM_RETURNED"
+            loan_dict["document_pid"] = "qaywb-gfe4B"
```
kprzerwa, line 338: minor: we could take this PID on top for readability like you did with item pids

cds_ils/cli.py
```diff
@@ -294,69 +292,94 @@ def create_demo_items(items_path):
...
-        end_date = start_date + timedelta(weeks=4)
-        item_pid = "678e3-an678A"
+    if total > 0 and not is_past_loan:
+        click.secho(
+            "Item for ongoing loan is already loaned by patron with email {0}."
+            .format(
+                active_loan[0].patron.email
+            ),
+            fg="red",
+        )
```
kprzerwa, line 316: you could just use `return` after this line and you wouldn't have to place the following code in the whole big `else:` block, that improves readability

cds_ils/cli.py
```diff
@@ -294,6 +292,32 @@ def create_demo_items(items_path):
 @with_appcontext
 def create_loan(user_email, is_past_loan):
     """Create a loan."""
+    ongoing_loan_item_pid = "vgrh9-jvj8E"
```
kprzerwa, line 295: please leave a comment that this part is hardcoded in demo_data because it is confusing from where the IDs are coming from

ntarocco (approved):

PR #225, opened 2020-11-10
vocab: Updates vocabularies values

jrcastro2:
Updates vocabularies values according to the librarians response (see [here](https://codimd.web.cern.ch/uR0k-4yiTzW0BSnp7mDw0Q?view#CDS-ILS-Vocabularies))


ntarocco (approved):
Thanks for this! We need to remember to update vocabs in the deployed envs.
There are already some CLI commands to do that, but changes of `keys` (which did not happen here) and deletion of `keys` (which did happen) should be handled manually (fix any existing record that has a referece to them)

@kprzerwa maybe a card in the board for the next deployment?

PR #227, opened 2020-11-10
migration: fix created_by field for documents

### LINKED ISSUE #154: Migration: fix _email import problem
in the field `created_by` we have a subfield `_email`.
There is no such field in the schema, the email needs to be translated for the user_id  or anonymous if the user does not exist in LDAP.
This action should be done on the import time.

- [x] replace default ingested medium type `NOT_SPECIFIED` -> `PAPER'
- [x] replace broken migrated values on the fields: `language`, `author.roles`
- [x] remove ed. from edition field on the migration time

KonstantinaStoikou:
Closes #154

### DIFF
--- a/cds_ils/migrator/utils.py
+++ b/cds_ils/migrator/utils.py
@@ -8,6 +8,7 @@
 
 """CDS Migrator Records utils."""
 from flask import current_app
+from invenio_accounts.models import User
 
 from cds_ils.migrator.errors import ItemMigrationError
 
@@ -129,7 +130,27 @@ def clean_item_record(record):
     clean_item_status(record)
     clean_description_field(record)
     record["shelf"] = record["location"]
-    record["medium"] = "NOT_SPECIFIED"
+    record["medium"] = "PAPER"  # requested as default value
     del record["location"]
     del record["id_bibrec"]
     del record["id_crcLIBRARY"]
+
+
+def clean_created_by_field(record):
+    """Clean the created by field for documents."""
+    if "_email" in record["created_by"]:
+        patron = User.query.filter_by(
+            email=record["created_by"]["_email"]
+        ).one_or_none()
+        if patron:
+            patron_pid = patron.get_id()
+            record["created_by"] = {"type": "user_id", "value": patron_pid}
+        else:
+            record["created_by"] = {"type": "user_id", "value": "anonymous"}
+        del record["created_by"]["_email"]
+    elif record["created_by"]["type"] == "batchuploader":
+        record["created_by"] = {"type": "script", "value": "batchuploader"}
+    else:
+        record["created_by"] = {"type": "script", "value": "migration"}
+
+    return record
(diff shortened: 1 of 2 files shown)


cds_ils/migrator/utils.py
```diff
@@ -129,7 +130,32 @@ def clean_item_record(record):
...
     del record["location"]
     del record["id_bibrec"]
     del record["id_crcLIBRARY"]
+
+
+def clean_created_by_field(record):
+    """Clean the created by field for documents."""
+    if "_email" in record["created_by"]:
+        patron = User.query.filter_by(
+            email=record["created_by"]["_email"]
```
ntarocco, line 143: Silly question: whenever I was trying this search, to filter by email field, the result was always empty. Does this really work?

cds_ils/migrator/utils.py
```diff
@@ -129,7 +130,32 @@ def clean_item_record(record):
...
+        ).one_or_none()
+        if patron:
+            patron_pid = patron.get_id()
+            record["created_by"]["type"] = "user_id"
+            record["created_by"]["value"] = patron_pid
+        else:
+            record["created_by"]["type"] = "user_id"
+            record["created_by"]["value"] = "anonymous"
+        del record["created_by"]["_email"]
+    elif record["created_by"]["type"] in [
```
ntarocco, line 153: Isn't it better to remove the extra check?
```suggestion
```
I think it might be better to always have a value, otherwise editing of migrated records might fail (not sure about this)?

cds_ils/migrator/utils.py
```diff
@@ -129,7 +130,32 @@ def clean_item_record(record):
     clean_item_status(record)
     clean_description_field(record)
     record["shelf"] = record["location"]
-    record["medium"] = "NOT_SPECIFIED"
+    record["medium"] = "PAPER"
```
kprzerwa, line 133: could you add a small comment like `# requested as default value`

cds_ils/migrator/utils.py
kprzerwa: @KonstantinaStoikou lets have a call about it because it does not cover all the possible cases we talked about

cds_ils/migrator/utils.py
KonstantinaStoikou: It worked but there were only 5 such records in the file I used.

ntarocco (approved):

cds_ils/migrator/utils.py
```diff
@@ -129,7 +130,31 @@ def clean_item_record(record):
...
+
+def clean_created_by_field(record):
+    """Clean the created by field for documents."""
+    if "_email" in record["created_by"]:
+        patron = User.query.filter_by(
+            email=record["created_by"]["_email"]
+        ).one_or_none()
+        if patron:
+            patron_pid = patron.get_id()
+            record["created_by"]["type"] = "user_id"
```
kprzerwa, line 147: in all the occurrences, to improve readability:
```
record["created_by"] = {"type": "user_id", "value" : patron_pid}
```

OR

```
record["created_by"] = dict(type="user_id", value=patron_pid)
```

PR #228, opened 2020-11-11
migration: mint legacy_recid [depends on react-invenio-app-ils#302]

### LINKED ISSUE #173: Mint legacy_recid. barcode, legacy_id as persistent ID

KonstantinaStoikou:
Closes #173 

- adds backend route `/api/legacy/<id>` that redirects to `api/documents/<pid>` or `api/series/<pid>`
- add frontend route `/legacy/<id>` that redirects to `/literature/<pid>` or `/series/<pid>`


kprzerwa (commented):
in general it is correct.

for the barcode, we should also mint it on the creation of the item itself, not only on migration time, because if there will be new items added later, we will not have their barcodes in the pidstore. Also the minters should be reusable, so they should be defined as a separate function.

How de we access a record by using legacy_recid? and the barcode?

kprzerwa (commented):
good job 🚀 
I left some minor comments to fix

ui/src/overridden/frontsite/Routes/LegacyRoute.js
```diff
@@ -0,0 +1,56 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      data: {},
+    };
+  }
+
+  componentDidMount() {
+    const { match } = this.props;
+    this.fetchDocumentFromLegacyRecId(match.params.legacyRecId);
```
kprzerwa, line 31: shouldn't we create a redux store for this, WDYT @ntarocco ?

kprzerwa:
there is a conflict, and requested changes in your PR from RIAILS

cds_ils/documents/api.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""CDS-ILS document API."""
+
+from invenio_app_ils.documents.api import Document as ILSDocument
+from invenio_pidstore.resolver import Resolver
+
+DOCUMENT_LEGACY_PID_TYPE = "lrecid"
+
+
+class Document(ILSDocument):
```
kprzerwa, line 16: I find it a bit dangerous in terms of naming, it is very easy to miss from where the Document was imported, CDS-ILS or invenio-app-ils. I would name it `CdsDocument` to leave no doubts.

KonstantinaStoikou: The same happens with Patron and AnonymousPatron class, they should be changed too? Also since we are using the `record_class` from the config do we care from where the class was imported?

cds_ils/minters.py
```diff
@@ -0,0 +1,23 @@
...
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""CDS-ILS minters."""
+
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+
+from cds_ils.documents.api import DOCUMENT_LEGACY_PID_TYPE
+
+
+def legacy_recid_minter(data, uuid):
```
kprzerwa, line 15: sorry for not noticing earlier:
you should pass `legacy_recid` as an argument, not in data dict, it wouldn't depend on a specific dictionary

kprzerwa (approved):

cds_ils/literature/views.py
```diff
@@ -0,0 +1,30 @@
...
+from invenio_records_rest.errors import PIDDoesNotExistRESTError
+
+legacy_blueprint = Blueprint("cds_ils_legacy", __name__)
+
+
+@legacy_blueprint.route("/legacy/<id>")
+def legacy_redirect(id):
+    """Redirect to the documents."""
+    try:
+        Document = current_app_ils.document_record_cls
```
ntarocco, line 22: Can it be that a legacy recid is a `series`?

cds_ils/migrator/records.py
```diff
@@ -275,8 +277,9 @@ def create_record(cls, dump):
         timestamp, json_data = dump.revisions[-1]
         json_data["pid"] = provider.pid.pid_value
         json_data = clean_created_by_field(json_data)
-
+        legacy_recid_minter(json_data["legacy_recid"], record_uuid)
```
ntarocco, line 280: we might need to do the same for series when importing?

setup.py
```diff
@@ -90,6 +90,7 @@
         ],
         "invenio_base.api_blueprints": [
             "cds_ils_patron_loans = cds_ils.patrons.views:create_patron_loans_blueprint",
+            "cds_ils_legacy = cds_ils.literature.views:legacy_blueprint",
```
ntarocco, line 93: missing to add the minter (see invenio-app-ils)
```
"invenio_pidstore.minters": [
```

ui/src/overridden/frontsite/Routes/LegacyRoute.js
ntarocco: I think in the past we said that it makes sense to created redux store/state when it is important to share state, or maybe I don't  recall correctly?

ntarocco (commented):
Nice implementation! Looks quite good to me, added a few comments.

I would propose to improve it a little bit more and I am sorry if I was not very clear when we had quick discussions about this.
In Invenio, you can mint a new PID as set is a redirected to another one (see [here an example](https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/models.py#L524)).
So what I would do is to mint it like that, with redirection, and then we resolving the legacy id should automatically redirect, removing some of this code. Please test it :)

The second point I have is that I am not sure I understand how `axios` is handling the redirection: is it "following" it? Because in the code I see a simple `.get(` and I am not sure it is fetching the Document record.

KonstantinaStoikou:
> The second point I have is that I am not sure I understand how `axios` is handling the redirection: is it "following" it? Because in the code I see a simple `.get(` and I am not sure it is fetching the Document record.

@ntarocco I think it is following the redirection because it does fetch the document.

cds_ils/series/api.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class Series(ILSSeries):
+    """CDS-Ils document class."""
+
+    @classmethod
+    def get_record_by_legacy_recid(cls, pid_value):
+        """Get ils record by pid value and pid type."""
+        resolver = Resolver(
+            pid_type=DOCUMENT_LEGACY_PID_TYPE,
```
kprzerwa, line 23:
```suggestion
            pid_type=SERIES_LEGACY_PID_TYPE,
```
or 
```suggestion
            pid_type=RECORD_LEGACY_PID_TYPE,
```
in general it would be better with `RECORD_LEGACY_PID_TYPE`, and maybe to move it into the config, since it is used by both Document and Series

setup.py
kprzerwa: added below

ui/src/overridableMapping.js
```diff
@@ -33,5 +34,6 @@ export const overriddenCmps = {
   'LoanMetadata.DeliveryIcon': DeliveryIcon,
   'RequestForPatronMessage.DeliveryIcon': DeliveryIcon,
   'LoanAvailability.NotAvailable': NotAvailable,
+  'Frontsite.route': LegacyDocumentRoute,
```
kprzerwa, line 37:
```suggestion
  'Frontsite.route': LegacyRecordRoute,
```
since it can be also series

ui/src/overridden/frontsite/Routes/LegacyRoute.js
kprzerwa:
```suggestion
    this.fetchRecordFromLegacyRecId(match.params.legacyRecId);
```
since it can be series too

cds_ils/literature/views.py
```diff
@@ -0,0 +1,32 @@
...
+        record = Document.get_record_by_legacy_recid(id)
+        if "document" in record["$schema"]:
+            return redirect("/api/documents/{0}".format(record.get("pid")))
```
kprzerwa, line 25: hardcoded URL

cds_ils/literature/views.py
```diff
@@ -0,0 +1,32 @@
...
+
+legacy_blueprint = Blueprint("cds_ils_legacy", __name__)
+
+
+@legacy_blueprint.route("/legacy/<id>")
+def legacy_redirect(id):
+    """Redirect to the documents."""
+    try:
+        Document = current_app_ils.document_record_cls
+        record = Document.get_record_by_legacy_recid(id)
```
ntarocco, line 23: The legacy recid can be a Document or a Series, in my opinion we should bind it to the Document.
I would change this to do:

```
from invenio_pidstore.resolver import Resolver
from invenio_records.api import Record

try:
    resolver = Resolver(
            pid_type=LEGACY_PID_TYPE, object_type="rec", getter=Record.get_record
        )
    _, record = resolver.resolve(str(pid_value))
except PersistentIdentifierError:
    return ...404...

is_document = schema.endswith(Document._schema)
is_series = schema.endswith(Series._schema)
assert is_document or is_series

url_path = "documents" if is_document else "series"

return redirect("/api/{0}/{1}".format(record["pid"], url_path))
```

cds_ils/documents/api.py
ntarocco, line 16: I think we should not have DOCUMENT_LEGACY_PID_TYPE and SERIES_LEGACY_PID_TYPE because they are identical.

We should probably not override `Document` and `Series` to add the method `get_record_by_legacy_recid`, but put that method as static util somewhere, probably in `literature/api.py`?:
```
LEGACY_PID_TYPE = "lrecid"

    def get_record_by_legacy_recid(cls, pid_value):
        """Get ils record by pid value and pid type."""
        resolver = Resolver(
            pid_type=LEGACY_PID_TYPE,
            object_type="rec",
            getter=cls.get_record,
        )
        _, record = resolver.resolve(str(pid_value))
        return record
```

cds_ils/minters.py
```diff
@@ -0,0 +1,23 @@
...
+
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+
+from cds_ils.documents.api import DOCUMENT_LEGACY_PID_TYPE
+
+
+def legacy_recid_minter(legacy_recid, uuid):
+    """Legacy_recid minter."""
+    PersistentIdentifier.create(
+        pid_type=DOCUMENT_LEGACY_PID_TYPE,
```
ntarocco, line 18:
```suggestion
        pid_type=LEGACY_PID_TYPE,
```

cds_ils/series/api.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: not needed

ui/src/api/legacy/legacy.js
```diff
@@ -0,0 +1,12 @@
+import { http } from '@inveniosoftware/react-invenio-app-ils';
+
+const legacyDocumentURL = '/legacy/';
```
ntarocco, line 3:
```suggestion
const legacyLiteratureURL = '/legacy/';
```

ui/src/overridableMapping.js
ntarocco: Literature or Record, find with me! Let's just try to keep it consistent everywhere, backend and frontend

ui/src/overridden/frontsite/Routes/LegacyRoute.js
```diff
@@ -0,0 +1,71 @@
...
+        component={LegacyDocumentCmp}
+      />
+      <Route>
+        <NotFound />
+      </Route>
+    </Switch>
+  );
+};
+
+export class LegacyDocumentCmp extends React.Component {
```
ntarocco, line 28:
```suggestion
export class Legacy[Record|Literature]Cmp extends React.Component {
```

ui/src/overridden/frontsite/Routes/LegacyRoute.js
```diff
@@ -0,0 +1,71 @@
...
+  withCancel,
+  FrontSiteRoutes,
+  recordToPidType,
+} from '@inveniosoftware/react-invenio-app-ils';
+import { Switch, Route, Redirect } from 'react-router-dom';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { legacyApi } from '../../../api/legacy/legacy';
+
+export const LegacyDocumentRoute = ({ ...props }) => {
```
ntarocco, line 13:
```suggestion
export const Legacy[Record|Literature]Route = ({ ...props }) => {
```

ui/src/overridden/frontsite/Routes/LegacyRoute.js
```diff
@@ -0,0 +1,71 @@
...
+
+  fetchDocumentFromLegacyRecId = async legacyRecId => {
+    this.cancellableFetchDocument = withCancel(legacyApi.get(legacyRecId));
+    const response = await this.cancellableFetchDocument.promise;
+    this.setState({ record: response.data });
+  };
+
+  render() {
+    const { record } = this.state;
+    console.log(record);
```
ntarocco, line 53:
```suggestion
```

cds_ils/migrator/items/api.py
```diff
@@ -51,10 +52,13 @@ def import_items_from_json(dump_file, include, rectype="item"):
 
                 # find document
                 try:
-                    record["document_pid"] = get_document_by_legacy_recid(
+                    Document = current_app_ils.document_record_cls
+                    record[
+                        "document_pid"
+                    ] = Document.get_record_by_legacy_recid(
```
ntarocco, line 58:
```suggestion
                    ] = get_record_by_legacy_recid(Document, record["id_bibrec"])
```

ntarocco (approved):
Great work! 👏

PR #229, opened 2020-11-12
covers: Fixes covers bug

jrcastro2:
Covers were displaying the Placeholder because the field `metadata.cover_metada.urls` was not present and in that case it will show the Placeholder component. 
Added to the serializer the missing field.

### DIFF
--- a/cds_ils/literature/serializers/json.py
+++ b/cds_ils/literature/serializers/json.py
@@ -7,6 +7,8 @@
 
 """Literature JSON serializers."""
 
+from invenio_app_ils.literature.serializers.custom_fields import \
+    field_cover_metadata
 from invenio_app_ils.literature.serializers.json import \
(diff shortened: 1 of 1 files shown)


cds_ils/literature/serializers/json.py
```diff
@@ -24,6 +26,7 @@ def transform_record(self, pid, record, links_factory=None, **kwargs):
         eitems = literature["metadata"].get("eitems", {}).get("hits", [])
         for eitem in eitems:
             format_login_required_urls(eitem)
+        field_cover_metadata(literature["metadata"])
```
ntarocco, line 29: oups it is me this one, sorry I didn't see it in `invenio-app-ils`, and I relied only on tests :(
Thanks for fixing it...

ntarocco (approved):

PR #236, opened 2020-11-23
Importer endpoints and logging

jrcastro2:
* addresses #213
* closes #215

PR #240, opened 2020-11-25
Importer: endpoints and ui

### LINKED ISSUE #213: Importer endpoint and UI interface
Backend: create a login/permission protected endpoint which expects a payload

`{
'provider': ..., 
'mode': ...
}`

And allows to upload a file. File should be validated as JSON or XML (check mimetype allowed in the REST endpoints). The endpoint should have a possibility to login with a token too (it should work out of the box on a invenio user account - you need to create one dedicated, needs only a test)

Providers can come from a config variable or from a vocabulary. Let's discuss briefly whats better.

UI:
* user should be able to choose a provider from the list.
* user should have a very simple interface to upload the file. File should be XML type and validates as such. Write it in a way to leave a possibility to upload JSON files in the future
* mode might be 'modify/create' or 'delete'. Ask the user for confirmation on the delete action

* on a success action (might be long, the status should be checked every 2 minutes) the user should be able to see the imported documents and items/eitems created (you can simulate it if the backend part not done yet)

The goal here is to have an entry point which allows to upload an XML file with records from providers, and it returns the links to imported records 

Bonus task (if we have time - otherwise please ticketize it)

- [ ] reverting the last action - if the librarian reviews the action and thinks that the whole thing was a mistake, we should be able to revert it. Caution: it requires accessing the record versions table, so lets evaluate it together

jrcastro2:
* closes #213
* ~depends on inveniosoftware/react-invenio-app-ils#308~ :heavy_check_mark: 

This new page allows us to import documents slecting the provider, the mode and uploading the file.

- All the fields are required. 
- The file estension must be `.xml`.
- If we select `delete` as mode, a confirmation modal will appear.
- While processing data (importing records), if we try to close/reload the results page we will get an alert.
- Each 30 seconds we will fetch the data (this can be changed).

Some screenshots:

### Importer form
image

### Form validation (frontend)
image

### Display of backend errrors on POST
image

### Confirmation modal on delete
image

### Loading
image

### Alert while loading on close/reload of tab
image

### Import failed
image

### Results page
image

## Invenio frontend

image

image

image

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -631,3 +631,7 @@ CDS_ILS_IMPORTER_PROVIDERS = {
     "ebl": {"priority": 3, "agency_code": "MiAaPQ"},
     "safari": {"priority": 4, "agency_code": "CaSebORM"},
 }
+
+CDS_ILS_IMPORTER_UPLOADS_PATH = "/tmp"
+
+CDS_ILS_IMPORTER_FILE_EXTENSIONS_ALLOWED = [".xml"]
--- a/cds_ils/importer/api.py
+++ b/cds_ils/importer/api.py
@@ -8,12 +8,12 @@
 """CDS-ILS Importer API module."""
 import logging
 
-import click
 from celery import shared_task
 from invenio_app_ils.errors import IlsValidationError
 from invenio_db import db
 
 from cds_ils.importer.errors import LossyConversion
+from cds_ils.importer.models import ImporterTaskEntry, ImporterTaskLog
 from cds_ils.importer.parse_xml import get_records_list
 from cds_ils.importer.XMLRecordLoader import XMLRecordDumpLoader
 from cds_ils.importer.XMLRecordToJson import XMLRecordToJson
@@ -48,48 +48,48 @@ def import_record(data, provider, source_type=None, eager=False):
         process_dump.delay(data, provider, source_type=source_type)
 
 
-def import_from_xml(sources, source_type, provider, eager=True):
-    """Load xml file."""
-    for idx, source in enumerate(sources, 1):
-        click.echo(
-            "({}/{}) Importing documents in {}...".format(
-                idx, len(sources), source.name
-            )
-        )
+def import_from_xml(log_id, source_path, source_type, provider):
+    """Load a single xml file."""
+    log = ImporterTaskLog.query.filter_by(id=log_id).first()
+    entry_data = None
+    try:
+        with open(source_path) as source:
+            records_list = list(get_records_list(source))
 
-        i = 0
-        for record in get_records_list(source):
-            try:
-                click.secho("Processing record {}".format(i))
-                i += 1
-                report = import_record(
-                    record, provider, source_type=source_type, eager=True
+            # update the entries count now that we know it
+            log.entries_count = len(records_list)
+            db.session.commit()
+
+            for i, record in enumerate(records_list):
+                entry_data = dict(
+                    import_id=log.id,
+                    entry_index=i,
                 )
-                click.secho(
-                    "Created: {}\n "
-                    "Updated: "
-                    "{}\n "
-                    "Ambiguous matches {}\n "
-                    "Fuzzy matches {}\n".format(
-                        report["created"],
-                        report["updated"],
-                        report["ambiguous_documents"],
-                        report["fuzzy"],
-                    ),
-                    fg="blue",
-                )
-            except LossyConversion as e:
-                continue
-            except IlsValidationError as e:
-                records_logger.error(
-                    "@FILE: {0} FATAL: {1}".format(
-                        source.name,
-                        str(e.original_exception.message),
+                try:
+                    report = import_record(record, provider,
+                                           source_type=source_type,
+                                           eager=True)
+                except LossyConversion as e:
+                    ImporterTaskEntry.create_failure(entry_data, e)
+                    continue
+                except IlsValidationError as e:
+                    records_logger.error(
+                        "@FILE TASK: {0} FATAL: {1}".format(
+                            log_id,
+                            str(e.original_exception.message),
+                        )
                     )
-                )
-                continue
-            except Exception as e:
-                records_logger.error(
-                    "@FILE: {0} ERROR: {1}".format(source.name, str(e))
-                )
-                raise e
+                    ImporterTaskEntry.create_failure(entry_data, e)
+                    continue
+
+                ImporterTaskEntry.create_success(entry_data, report)
+    except Exception as e:
+        records_logger.error(
+            "@FILE TASK: {0} ERROR: {1}".format(log_id, str(e))
+        )
+        if entry_data:
+            ImporterTaskEntry.create_failure(entry_data, e)
+        log.set_failed(e)
+        raise e
+
+    log.set_succeeded()
--- a/cds_ils/importer/views.py
+++ b/cds_ils/importer/views.py
@@ -1,0 +1,171 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Views for importer."""
+
+import json
+import os
+import uuid
+from threading import Thread
+
+import arrow
+from flask import Blueprint, abort, copy_current_request_context, request
+from invenio_app_ils.permissions import need_permissions
+from webargs import fields
+from webargs.flaskparser import use_kwargs
+
+from cds_ils.importer.api import import_from_xml
+from cds_ils.importer.models import ImporterAgent, ImporterMode, \
+    ImporterTaskEntry, ImporterTaskLog
+
+
+def create_importer_blueprint(app):
+    """Add importer views to the blueprint."""
+    blueprint = Blueprint("invenio_app_ils_importer", __name__)
+
+    def allowed_files(filename):
+        """Checks the extension of the files."""
+        allowed_extensions = app.config[
+            "CDS_ILS_IMPORTER_FILE_EXTENSIONS_ALLOWED"
+        ]
+        return filename.lower().endswith(tuple(allowed_extensions))
+
+    def rename_file(filename):
+        """Renames filename with an unique name."""
+        unique_filename = uuid.uuid4().hex
+        ext = filename.rsplit(".", 1)[1]
+        return unique_filename + "." + ext
+
+    @blueprint.errorhandler(413)
+    def payload_too_large(error):
+        return {"message": "The file is too large", "status": 413}, 413
+
+    @blueprint.route("/importer/check/<int:log_id>", methods=["GET"])
+    @need_permissions("document-importer")
+    def check(log_id):
+        return check_next(log_id, 0)
+
+    @blueprint.route("/importer/check/<int:log_id>/next/<int:next_entry>",
+                     methods=["GET"])
+    @need_permissions("document-importer")
+    def check_next(log_id, next_entry):
+        log = ImporterTaskLog.query.filter_by(id=log_id).first()
+        if log:
+            children_entries_query = ImporterTaskEntry.query \
+                .filter_by(import_id=log_id)
+            entries = children_entries_query \
+                .filter(ImporterTaskEntry.entry_index >= next_entry) \
+                .order_by(ImporterTaskEntry.entry_index.asc()) \
+                .all()
+            obj = {
+                "id": log_id,
+                "state": log.status.value,  # enum value
+                "start_time": arrow.get(log.start_time).isoformat(),
+                "end_time": arrow.get(log.end_time).isoformat() \
+                if log.end_time else None,
+                "original_filename": log.original_filename,
+                "provider": log.provider,  # string
+                "mode": log.mode.value,  # enum value
+                "source_type": log.source_type,  # string
+            }
+            if log.entries_count:
+                obj["total_entries"] = log.entries_count
+                obj["loaded_entries"] = children_entries_query.count()
+            reports = []
+            for entry in entries:
+                if not entry.error:
+                    reports.append({
+                        "index": entry.entry_index,
+                        "success": True,
+                        "report": {
+                            "ambiguous_documents": entry.ambiguous_documents,
+                            "ambiguous_eitems": entry.ambiguous_eitems,
+                            "created_document": entry.created_document,
+                            "created_eitem": entry.created_eitem,
+                            "updated_document": entry.updated_document,
+                            "updated_eitem": entry.updated_eitem,
+                            "deleted_eitems": entry.deleted_eitems,
+                            "series": entry.series,
+                            "fuzzy_documents": entry.fuzzy_documents,
+                        }
+                    })
+                else:
+                    reports.append({
+                        "index": entry.entry_index,
+                        "success": False,
+                        "message": entry.error,
+                    })
+            obj["reports"] = reports
+            return obj
+        else:
+            return 404
+
+    @blueprint.route("/importer", methods=["POST"])
+    @use_kwargs({"provider": fields.Str(required=True)})
+    @use_kwargs({"mode": fields.Str(required=True)})
+    @need_permissions("document-importer")
+    def importer(provider, mode):
+        @copy_current_request_context
+        def import_from_xml_background(log_id, file, source_type, provider):
+            """Acts as a proxy to pass the current context to the function."""
+            import_from_xml(log_id, file, source_type, provider)
+
+        def create_import_task(source_path, original_filename, source_type,
+                               provider, mode):
+            """Creates a task and returns its associated identifier."""
+            importer_mode_map = dict(
+                create=ImporterMode.CREATE,
+                delete=ImporterMode.DELETE,
+            )
+
+            log = ImporterTaskLog.create(dict(
+                agent=ImporterAgent.USER,
+                provider=provider,
+                source_type=source_type,
+                mode=importer_mode_map[mode],
+                original_filename=original_filename,
+            ))
+
+            t = Thread(target=import_from_xml_background,
+                       args=(log.id, source_path, source_type, provider))
+            t.start()
+
+            return log.id
+
+        if request.files:
+            file = request.files["file"]
+
+            if not provider:
+                abort(400, "Missing provider")
+
+            if not mode:
+                abort(400, "Missing mode")
+
+            if allowed_files(file.filename):
+                original_filename = file.filename
+                file.filename = rename_file(file.filename)
+                source_path = os.path.join(
+                    app.config["CDS_ILS_IMPORTER_UPLOADS_PATH"],
+                    file.filename
+                )
+                file.save(source_path)
+
+                source_type = "marcxml"
+                log_id = create_import_task(source_path,
+                                            original_filename,
+                                            source_type,
+                                            provider, mode)
+
+                return (json.dumps({"id": log_id}),
+                        201,
+                        {"ContentType": "application/json"})
+            else:
+                abort(400, "That file extension is not allowed")
+        else:
+            abort(400, "Missing file")
+
+    return blueprint
--- a/cds_ils/ldap/tasks.py
+++ b/cds_ils/ldap/tasks.py
@@ -11,6 +11,7 @@
 from celery import shared_task
 from celery.utils.log import get_task_logger
 from flask import current_app
+from invenio_db import db
 
 from .api import update_users
 from .models import LdapSynchronizationLog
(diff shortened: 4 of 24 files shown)


jrcastro2:
Note for reviewers: Please just review last commit (the UI changes), the backend changes can be checked [here](https://github.com/CERNDocumentServer/cds-ils/pull/236).

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,189 @@
...
+                      active={activeIndex === index}
+                      index={index}
+                      onClick={this.handleClick}
+                      color="red"
+                    >
+                      <Icon name="dropdown" />
+                      {!_isEmpty(document) ? (
+                        <Link
+                          to={BackOfficeRoutes.documentDetailsFor(document.pid)}
+                          target="_blank"
```
FlorianCassayre, line 159: 👍  nice detail

cds_ils/patrons/permissions.py
```diff
@@ -25,9 +25,17 @@ def retrieve_patron_loans_permission(*args, **kwargs):
     )
 
 
+def backoffice_permission(*args, **kwargs):
+    """Return permission to allow only librarians and admins."""
+    return Permission(backoffice_access_action)
```
FlorianCassayre, line 30: Hm, why do we need to redefine `backoffice_permission`? (there is one in ILS already)

FlorianCassayre (commented):
Good job, very nice UI!
A couple of minor remarks:
* the first progress update is important because it contains valuable informations (how many entries have to be processed) but it takes some time to receive it (30 seconds currently)
* the top border appears to be missing:
image

cds_ils/patrons/permissions.py
jrcastro2: Yes, you are right this code is lefotver from a bad rebase, thanks!

FlorianCassayre: Okay! I've implemented the importer permissions in that file already but maybe it's not the best place. In any case this will be done in the other PR.

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                else:
+                    reports.append({
+                        "index": entry.entry_index,
+                        "success": False,
+                        "message": entry.error,
+                    })
+            obj["reports"] = reports
+            return obj
+        else:
+            raise Exception("Task not found")  # TODO
```
ntarocco, line 113:
```suggestion
            return 404
```

If you want to keep it cleaner, maybe return at the top:
```
if not log:
    return 404
```
or extract:
```
if log:
    return get_report(log)
else:
    return 404
```

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                t.start()
+
+                return (json.dumps({"id": log.id}),
+                        200,
+                        {'ContentType': 'application/json'})
+            else:
+                flash('That file extension is not allowed')
+                raise RequestError('That file extension is not allowed', 400)
+        else:
+            raise RequestError('Missing file', 400)
```
ntarocco, line 167: I would replace all `raise`, in a view/request context, with `abort`.

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+            obj["reports"] = reports
+            return obj
+        else:
+            raise Exception("Task not found")  # TODO
+
+    @blueprint.route('/importer', methods=['POST'])
+    @need_permissions("document-importer")
+    def importer():
+
+        @copy_current_request_context
```
ntarocco, line 119: why is this needed?

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                raise RequestError('Missing mode', 400)
+
+            if allowed_files(file.filename):
+                original_filename = file.filename
+                file.filename = rename_file(file.filename)
+                filename = secure_filename(file.filename)
+                file.save(os.path.join(app.config["IMPORTER_UPLOADS_PATH"],
+                                       filename))
+
+                # TODO move this to API
```
ntarocco, line 145: TODO warning

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                log = ImporterTaskLog.create(dict(
+                    agent=ImporterAgent.USER,
+                    provider=provider,
+                    source_type=source_type,
+                    mode=ImporterMode.CREATE if mode == "create"
+                    else ImporterMode.UPDATE,
+                    original_filename=original_filename,
+                ))
+
+                t = Thread(target=process_data,
```
ntarocco, line 156: why a new thread in a request context? What is the advantage?

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                        "message": entry.error,
+                    })
+            obj["reports"] = reports
+            return obj
+        else:
+            raise Exception("Task not found")  # TODO
+
+    @blueprint.route('/importer', methods=['POST'])
+    @need_permissions("document-importer")
+    def importer():
```
ntarocco, line 117: can we add a max size validation?

ntarocco, line 117: did you try to upload a `bla.xml` which contain a non valid XML? Everything good?

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                      + file.filename) as f:
+                import_from_xml(log_id, f, source_type, provider)
+
+        if request.files:
+            file = request.files["file"]
+            provider = request.form.get("provider", None)
+            mode = request.form.get("mode", None)
+
+            if not provider:
+                flash('Missing provider')
```
ntarocco, line 131: isn't `flash` used only in Jinja templates? Why is it needed?

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+
+        @copy_current_request_context
+        def process_data(log_id, file, source_type, provider):
+            with open(app.config["IMPORTER_UPLOADS_PATH"] + "/"
+                      + file.filename) as f:
+                import_from_xml(log_id, f, source_type, provider)
+
+        if request.files:
+            file = request.files["file"]
+            provider = request.form.get("provider", None)
```
ntarocco, line 127: Please prefer `webargs` instead, you have args validation and defaults out of the box.

cds_ils/importer/views.py
```diff
@@ -0,0 +1,169 @@
...
+                raise RequestError('Missing provider', 400)
+
+            if not mode:
+                flash('Missing mode')
+                raise RequestError('Missing mode', 400)
+
+            if allowed_files(file.filename):
+                original_filename = file.filename
+                file.filename = rename_file(file.filename)
+                filename = secure_filename(file.filename)
```
ntarocco, line 141: renaming 2 times? Why is that?

cds_ils/ldap/tasks.py
```diff
@@ -28,5 +29,6 @@ def synchronize_users_task():
         result = update_users()
         log.set_succeeded(*result)
     except Exception as e:
+        db.session.rollback()
```
ntarocco, line 32: 👍 great! Were you able to validate that it actually fixes the issue?

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+      activeIndex: -1,
+      importCompleted: false,
+      data: null,
+      intervalId: null,
+      isLoading: true,
+    };
+  }
+
+  componentDidMount() {
+    //TODO: Define how much we wait until the next call (currently 30 sec)
```
ntarocco, line 27: TODO warning. Why not using a config instead?
30 secs looks a lot to me... Then I would expect to see something like: checking in 30... 29... etc...

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+    const idVar = setInterval(() => this.checkForData(idVar), 30000);
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ intervalId: idVar });
```
ntarocco, line 30: No need for React state here:
```suggestion
    this.intervalId = idVar;
```

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+    this.checkForData(idVar);
+    window.addEventListener('beforeunload', this.onCloseWarning);
+  }
+
+  componentWillUnmount = () => {
+    const { intervalId } = this.state;
+    window.removeEventListener('beforeunload', this.onCloseWarning);
+    clearInterval(intervalId);
```
ntarocco, line 38: Safer to check if defined:
```suggestion
    this.intervalId && clearInterval(this.intervalId);
```

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+    const idVar = setInterval(() => this.checkForData(idVar), 30000);
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ intervalId: idVar });
+    this.checkForData(idVar);
+    window.addEventListener('beforeunload', this.onCloseWarning);
+  }
+
+  componentWillUnmount = () => {
+    const { intervalId } = this.state;
+    window.removeEventListener('beforeunload', this.onCloseWarning);
```
ntarocco, line 37: Does it failed if not defined for some reasons? Otherwise:
```
this.onCloseWarning && window.removeEventListener('beforeunload', this.onCloseWarning);
```

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+    clearInterval(intervalId);
+  };
+
+  onCloseWarning = e => {
+    const { isLoading } = this.state;
+    if (isLoading) {
+      var confirmationMessage =
+        'Importing in progress. Are you sure you want to leave?';
```
ntarocco, line 45: What is happening if I leave? Everything lost? Then better to warn the user:
`Importing in progress. If you leave the page, the import will be cancelled. Are you sure?`

ntarocco (commented):
One of the main comment is that I have the feeling that if it is a long task, we should this in a celery task so that if I leave the web page, the process is not lost.
There are a lot of things that can go wrong just waiting on a webpage: loosing connection, closing the laptop lid, the battery of the laptop dies... :) this is one of the biggest issue with have in CDS Videos.
Maybe to be discussed in IRL.

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+    }
+  };
+
+  checkForData = async idVar => {
+    const { importCompleted } = this.state;
+    const { taskId } = this.props;
+
+    if (!importCompleted) {
+      // eslint-disable-next-line no-useless-concat
+      const response = await http.get(`${importerURL}` + '/' + `${taskId}`);
```
ntarocco, line 58: I would have maybe created a nice endpoint in the `api.py` as for the others, instead of having it inside a component.

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+          </>
+        ) : data.state === 'SUCCEEDED' ? (
+          <>
+            <Icon name="check" />
+            Documents imported succesfully.
+          </>
+        ) : (
+          <>
+            <Icon name="close" />
+            Documents import failed.
```
ntarocco, line 114: It can be also Series and EItems? If yes, I would use Literature everywhere.

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
```diff
@@ -0,0 +1,200 @@
...
+                    ' documents out of ' +
+                    data.total_entries +
+                    '.'}
+                </p>
+              ) : (
+                <p>{'Processed ' + data.reports.length + ' documents.'}</p>
+              )
+            ) : null}
+            {!_isEmpty(data.reports) ? (
+              <Accordion className="importer" styled fluid>
```
ntarocco, line 146: can we refactor and extract in smaller cmps?

ui/src/overridden/backoffice/Importer/Importer.js
```diff
@@ -0,0 +1,253 @@
...
+
+  postData = async formData => {
+    const importerURL = '/importer';
+    const headers = {
+      headers: {
+        'Content-Type': 'multipart/form-data',
+      },
+    };
+    try {
+      const response = await http.post(`${importerURL}`, formData, headers);
```
ntarocco, line 53: same here for the endpoint, to have a module in `api` to have all endpoints together.

ui/src/overridden/backoffice/Importer/Importer.js
```diff
@@ -0,0 +1,253 @@
...
+      this.setState({
+        error: error,
+      });
+    }
+  };
+
+  handleSubmit = async () => {
+    const { provider, mode, file } = this.state;
+    if (!_isEmpty(provider) && !_isEmpty(mode) && file) {
+      var formData = new FormData();
```
ntarocco, line 72: we are using a React upload component to upload files for eitems, why not re-using that one and get rid of custom code?

ui/src/overridden/backoffice/Importer/Importer.js
```diff
@@ -0,0 +1,253 @@
...
+      </Form>
+    );
+  };
+
+  render() {
+    const { postDone, error } = this.state;
+    return (
+      <Container id="importer" className="spaced">
+        <>
+          <Header as="h1">Documents Importer</Header>
```
ntarocco, line 241:
```suggestion
          <Header as="h1">Literatures Importer</Header>
```

cds_ils/importer/views.py
jrcastro2: Needed for the thread to "work" inside the view function.

cds_ils/importer/views.py
jrcastro2: It's used to launch the importing process and to be able to return a 200 (if everything ok) at the same time. Otherwise we would be waiting for the response

cds_ils/importer/views.py
jrcastro2: Yes, an error warning you that the import failed will appear.

cds_ils/importer/views.py
jrcastro2: Yes, will remove it, it's leftover code, sorry and thanks!

cds_ils/importer/views.py
jrcastro2: You are right, the second one is no longer needed since we have unique names.

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
jrcastro2: It doesn't fail but I will add it, just in case

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
jrcastro2: No, if you leave you will not be able to keep track of the import, but it will keep running until it succeeds or fails.

ui/src/overridden/backoffice/Importer/ImportedDocuments.js
jrcastro2: If I am not wrong we import documents, but we also create/update/delete eitems related to them, so yes we can use Literature

FlorianCassayre:
Some of the comments are for the other PR, I'll take care of them.

ui/src/overridden/backoffice/Importer/Importer.js
jrcastro2: Hmm, if you mean this component I did not use it because it seems to specific for the eitems, it needs an eitem as a prop, it launches the call on selecting the file, while we only want to store it and send it along with the rest of the data from the form. Maybe I can create a more generic component re-usable by both use cases

cds_ils/importer/views.py
jrcastro2: Added, but maybe the current max size is too much ? (100 MiB) (see [here](https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/config.py#L186))

cds_ils/ldap/tasks.py
FlorianCassayre: I tested it this morning and it was in fact the cause of the issue of the tasks being left "RUNNING". We had a test that covered failures, but database errors are different in the sense that they have side-effects (and we must rollback the changes when one occurs, before committing further changes). The errors are now correctly reported in the database.

cds_ils/importer/views.py
FlorianCassayre: Hmm maybe, the main danger is that we are running it on the webserver node, so we must be careful not to load too much data in memory.

cds_ils/importer/api.py
```diff
@@ -48,48 +48,39 @@ def import_record(data, provider, source_type=None, eager=False):
...
-    """Load xml file."""
-    for idx, source in enumerate(sources, 1):
-        click.echo(
-            "({}/{}) Importing documents in {}...".format(
-                idx, len(sources), source.name
-            )
-        )
+def import_from_xml(log_id, source, source_type, provider):
+    """Load a singular xml file."""
+    log = ImporterTaskLog.query.filter_by(id=log_id).first()
```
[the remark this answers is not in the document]
FlorianCassayre: I was encountering issues when changing context if I passed the database object. I'll see what I can do.

cds_ils/config.py
```diff
@@ -631,3 +631,7 @@ def _parse_env_bool(var_name, default=None):
     "ebl": {"priority": 3, "agency_code": "MiAaPQ"},
     "safari": {"priority": 4, "agency_code": "CaSebORM"},
 }
+
+CDS_ILS_IMPORTER_UPLOADS_PATH = "/tmp"
+
+CDS_ILS_IMPORTER_FILE_EXTENSIONS_ALLOWED = ["xml"]
```
ntarocco, line 637: Very minir: no `.` in the end? Otherwise I could probably upload something `.axml` :)

cds_ils/importer/views.py
```diff
@@ -0,0 +1,171 @@
...
+        """Renames filename with an unique name."""
+        unique_filename = uuid.uuid4().hex
+        ext = filename.rsplit(".", 1)[1]
+        return unique_filename + "." + ext
+
+    @blueprint.errorhandler(413)
+    def payload_too_large(error):
+        return {"message": "The file is too large", "status": 413}, 413
+
+    @blueprint.route("/importer/check/<int:log_id>", methods=["GET"])
```
ntarocco, line 47: why not simply:
- `/importer/tasks/` for the list
- `/importer/tasks/<job_id>` for the details
(probably to be added to the new GH issue)

cds_ils/importer/views.py
```diff
@@ -0,0 +1,171 @@
...
+    @need_permissions("document-importer")
+    def check(log_id):
+        return check_next(log_id, 0)
+
+    @blueprint.route("/importer/check/<int:log_id>/next/<int:next_entry>",
+                     methods=["GET"])
+    @need_permissions("document-importer")
```
ntarocco, line 54: can you explain why the `next` param? Is it like a range? Why is it needed?

cds_ils/importer/views.py
```diff
@@ -0,0 +1,171 @@
...
+                        "index": entry.entry_index,
+                        "success": False,
+                        "message": entry.error,
+                    })
+            obj["reports"] = reports
+            return obj
+        else:
+            return 404
+
+    @blueprint.route("/importer", methods=["POST"])
```
ntarocco, line 107:
```suggestion
    @blueprint.route("/importer/tasks", methods=["POST"])
```
to be added to the ticket?

cds_ils/importer/views.py
```diff
@@ -0,0 +1,171 @@
...
+                file.save(source_path)
+
+                source_type = "marcxml"
+                log_id = create_import_task(source_path,
+                                            original_filename,
+                                            source_type,
+                                            provider, mode)
+
+                return (json.dumps({"id": log_id}),
+                        200,
```
ntarocco, line 164:
```suggestion
                        201,
```

ntarocco (approved):

cds_ils/config.py
FlorianCassayre: Nice catch, you're totally right.

cds_ils/importer/views.py
FlorianCassayre: Should have been documented; the idea is to avoid fetching all the entries each time because they don't change, and instead only ask for the ones that we don't have. Reason is that the response payload size grows linearly otherwise (and fast!).

PR #249, opened 2020-11-30
ldap: disable error on missing email

ntarocco:


FlorianCassayre (approved):
Great, I was wondering if we could make this assumption.
Keep in mind that there is still an issue with some emails being assigned to new employee IDs.

**Edit**: discussed about it

kprzerwa (approved):

PR #252, opened 2020-12-13
ldap: fix import on duplicated emails

### LINKED ISSUE #239: LDAP user synchronization
There are two "simple" problems remaining:
* first one, major, the synchronization can still fail in a specific case
  * minimal example scenario:
    * invenio users is initially made of `[(1, a@test.ch)]` (format is person id, email)
    * now for whatever reason ldap returns `[(2, a@test.ch)]`
    * because we are matching on person ids, the script will consider this particular ldap user as being a new invenio user and will try to create it. That's when it will fail because the email column has a unique key constraint (for good reasons)
  * I'm not sure why this would happen, but it did. I would say we do not create a new user and completely ignore it if the email is already associated to an active invenio account
* second one, in case of a database error the exception message is not reported in the database and the entry is left as "RUNNING". This is likely due to the fact that the transaction is not explicitly rollbacked in case of a failure

Both issues can be covered by tests. As it turns out the small example scenario should be enough actually.

ntarocco:
* closes #239

For reviewers: the code changes are only a refactoring to spit that huge function in smaller function (that could be extracted in a class later). I have highlighted the only change that I have added with a comment.

### DIFF
--- a/cds_ils/ldap/api.py
+++ b/cds_ils/ldap/api.py
@@ -463,15 +463,15 @@ def update_users():
     invenio_users = remap_invenio_users(log_func)
 
     # STEP 1 - update Invenio users with info from LDAP
-    ldap_users_map, invenio_users_updated_count = update_invenio_users_from_ldap(
+    ldap_users_map, invenio_users_updated = update_invenio_users_from_ldap(
         invenio_users, ldap_users_map, log_func
     )
 
     # STEP 2 - import any new LDAP user not in Invenio yet
-    invenio_users_added_count = 0
+    invenio_users_added = 0
     new_ldap_users = ldap_users_map.values()
     if new_ldap_users:
-        invenio_users_added_count = import_new_ldap_users(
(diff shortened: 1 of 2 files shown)


cds_ils/ldap/api.py
```diff
@@ -417,54 +402,95 @@ def update_users():
...
+        """Import any new LDAP user not in Invenio yet."""
         importer = LdapUserImporter()
+        added_count = 0
         for ldap_user in new_ldap_users:
+            # Check if email already exists in Invenio.
+            # Apparently, in some cases, there could be multiple LDAP users
+            # with different person id but same email.
+            email = ldap_user_get_email(ldap_user)
+            email_exists = User.query.filter_by(email=email).count() > 0
+            if email_exists:
```
ntarocco, line 422: This is the real fix, to skip duplicated email

jrcastro2 (approved):

kprzerwa (approved):

FlorianCassayre (approved):

PR #244, opened 2020-11-27
[depends on #240] Importer deletion

jrcastro2:
Allows deletion mode on importer

PR #261, opened 2021-01-06
importer: enhancements

### LINKED ISSUE #245: importer: enhancements
Following the demo of #240, the following suggestions were proposed:
* [x] the per-entry errors are not explicit enough:
  * in case of a failure, information about the literature should be displayed in order to identify it (title, ...). Example of expected error: `The literature <title or other fields> was not imported. Error: LossyRule 980__a`
  * the error should explicitly say if the literature was imported, but it requires curation, or it was not imported at all.
* in the frontend:
  * [x] the fields "created", "updated", etc. are not self-explanatory. Try to improve the user experience (One suggestion was to create 4 "sections", something like `Documents | Series | Eitems | Error`)
  * [x] Add icons (e.g. green tick/red cross) to the importer results page to be able to quickly distinguish between succeeded, failed and no update/created records.
  * [x] transform the view to a class list/details page:
    * the list `/tasks/` should contain the import and the current/previous jobs. When clicking on "import", the view should redirect to the job page `/tasks/<taskid>` to see the status. The upload form should be blocked. No need for search / pagination. 
    * the task page `/tasks/<taskid>` should display what is currently displayed, dates, what was imported and errors.
* the tasks are executed from the webserver node on a parallel thread, because they depend on a file that was uploaded there. To do:
  * [x] move the work from being done in a thread, to a celery shared task (at this point one can test it locally)
  * [ ] create a shared volume on OpenShift between the webserver and celery worker nodes, override the env config value to set its path to the shared volume
* in invenio admin there are currently two pages accessible from the sidebar, one for each relation (tasks + imported literature). The relations are linked in a one-to-many fashion but this link is not shown in a user-friendly way. To do:
  * [x] the sidebar should only display the link to the tasks relation
  * [x] each task row should have a link that lists all the entries (for that task), similarly to what is done for files/objects
* [x] importer logic: if a document has eitems already created, it would fail if the `created_by` value does not match `import:xxx` (seems to have been fixed already)

FlorianCassayre:
* ~depends on inveniosoftware/react-invenio-app-ils#323 (requires the `ResultsTable` component)~
  * requires `react-invenio-app-ils v1.0.0-alpha.21`
* closes #245
* addresses comments in #244

Summary of changes:
* the tasks are now executed by the celery workers, instead of the web server nodes. The config value `CDS_ILS_IMPORTER_UPLOADS_PATH` should point to a partition that is available to both the webserver and the celery worker (see [merge request 8 on gitlab](https://gitlab.cern.ch/cds-team/cds-ils-openshift/-/merge_requests/8))
* each task has its own backoffice URL (`/importer/tasks/:id`); when the form is submitted the user is simply redirected there
* added an endpoint `/importer/tasks/list` listing the latest imports, and integrated the data in the UI (right below the form)
* improved invenio admin panel: only the task relation is shown in the menu, but the user has the ability to view all the event entries for any given task (similar to the file buckets)
* improved display of data and errors
* codestyle & refactor

image

---

image

---

image

---

image

---

The relation below can be accessed by clicking on the "Record updates" link. It will filter the rows that are children of the clicked task (here **3**).

image


ui/src/overridden/backoffice/Importer/ImporterTaskDetails/ReportDetails.js
```diff
@@ -0,0 +1,189 @@
...
+    );
+  };
+
+  render() {
+    const { item } = this.props;
+    return (
+      <Item>
+        <Item.Content>
+          {item.success ? (
+            <Table definition compact striped textAlign="center">
```
kprzerwa, line 57: there is a table component which is more compact in terms of code and could be used for this

ui/src/overridden/backoffice/Importer/ImporterTaskDetails/ReportDetails.js
```diff
@@ -0,0 +1,189 @@
...
+        </span>
+      ));
+};
+
+export class ReportDetails extends React.Component {
+  renderErrorMessage = message => {
+    let title;
+    let content;
+    const split = message.split(': ');
+    if (split.length > 1 && split[0] === 'LossyConversion') {
```
kprzerwa, line 36: the error message normally should be provided by the backend, instead of custom parsing it's text on the frontend

kprzerwa (commented):
looks good, I left some comments
we should demo it on Monday!

ui/src/overridden/backoffice/Importer/ImporterTaskDetails/ReportDetails.js
FlorianCassayre: Currently we format the errors that way so they can be both human readable and parsable without ambiguity. I'm not sure if we would gain much by structuring the data (in json for instance), data which consists of just two fields.

FlorianCassayre:
Applied the requested changes, also fixed the issue encountered in the demo - which was a typo. Successfully tested & updated the openshift config, **dev only** (will have to propagate the changes to sandbox once we validate this, after deployment). That means it should already be working on the next deployment.

ui/src/overridden/backoffice/Importer/Importer.js
```diff
@@ -46,15 +214,10 @@ export default class Importer extends Component {
...
-      this.setState({
-        provider: '',
-        mode: '',
-        file: null,
-        postDone: true,
-        taskId: response.data.id,
-        error: {},
-      });
+      const taskId = response.data.id;
+      history.push(BackOfficeRouteGenerators.importerDetailsFor(taskId));
```
ntarocco, line 218:
```suggestion
      goTo(BackOfficeRouteGenerators.importerDetailsFor(taskId));
```

ntarocco (approved):

ui/src/overridden/backoffice/Importer/Importer.js
```diff
@@ -9,14 +10,183 @@ import {
...
+  datetimeFormatter = ({ col, row }) => {
+    const datetime = DateTime.fromISO(row[col.field]);
+    return datetime.toLocaleString(
+      Object.assign(DateTime.DATETIME_SHORT, { locale: 'en-GB' })
+    );
+  };
+
+  durationFormatter = ({ col, row }) => {
+    const endTime = row[col.field];
+    if (endTime) {
```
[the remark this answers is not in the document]
FlorianCassayre: Technically Luxon is already providing most of the logic, the only thing left to do is to print the duration without leading zeros and units pluralized. Maybe I over-engineered it...

ui/src/overridden/backoffice/Importer/Importer.js
FlorianCassayre: Hm for this one the reason I couldn't use `goTo` is because it was not exported!

FlorianCassayre:
Up and running on dev: [https://dev-cds-ils.web.cern.ch/backoffice/importer]()

ntarocco (approved):

PR #262, opened 2021-01-07
patrons: index patrons only if they have remote account

### LINKED ISSUE #246: EDH termination sheet: do not index the local user
For the EDH termination sheet, a new Invenio system local user is created, but this should be not indexed in ES and be visible anywhere.
One solution could be to create a config to blacklist the accounts to index.
Another solution: do not index the users without connected remote account.

KonstantinaStoikou:
- override `reindex_patrons` function with function that also checks if user has a remote account
- add configuration to whether local accounts should be indexed or not

Depends on inveniosoftware/invenio-app-ils#1012
Closes #246

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -47,6 +47,7 @@ from marshmallow.fields import Bool, List
 from .circulation.utils import circulation_cds_extension_max_count
 from .literature.covers import build_cover_urls
 from .patrons.api import AnonymousPatron, Patron
+from .patrons.indexer import PatronIndexer
 from .patrons.permissions import views_permissions_factory
 
 
@@ -360,6 +361,7 @@ JSONSCHEMAS_SCHEMAS = [
 # RECORDS REST
 ###############################################################################
 RECORDS_REST_ENDPOINTS[PATRON_PID_TYPE]["record_class"] = Patron
+RECORDS_REST_ENDPOINTS[PATRON_PID_TYPE]["indexer_class"] = PatronIndexer
 RECORDS_REST_ENDPOINTS[LOCATION_PID_TYPE][
     "create_permission_factory_imp"
 ] = deny_all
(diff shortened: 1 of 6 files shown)


cds_ils/patrons/indexer.py
```diff
@@ -0,0 +1,37 @@
...
+    all_user_ids = db.session.query(User.id).all()
+    indexed_count = 0
+    for (user_id,) in all_user_ids:
+        if (
+            LOCAL_ACCOUNTS_INDEXED
+            or RemoteAccount.query.filter(
+                RemoteAccount.user_id == user_id
+            ).first()
+            is not None
+        ):
```
kprzerwa, line 33: for readability:
```suggestion
     index_local_account = current_app.config['INDEX_LOCAL_ACCOUNTS')
     has_remote_account = RemoteAccount.query.filter_by(user_id=user_id).first()
     if index_local_account or has_remote_account:
```

other: 
in the `or` statement `None` will be evaluated as false, there is no need for additional check `is not None`

another optimisation:
wouldn't it be better if :
```python
if index_local_account:
    all_user_ids = db.session.query(User.id).all()
else:
    all_user_ids = db.session.query(RemoteAccount.user_id).all()
```

in this case in each option you make only one db query not two

kprzerwa (commented):
LGTM 🚀 
 I would say only small unit test missing

kprzerwa (approved):

setup.py
```diff
@@ -78,6 +78,7 @@
             "importer = cds_ils.importer.cli:importer",
             "fixtures = cds_ils.cli:fixtures",
             "covers = cds_ils.cli:covers",
+            "patrons = cds_ils.patrons.cli:patrons",
```
kprzerwa, line 81: is there a name conflict when registering this command? since it is the same in app-ils

KonstantinaStoikou: you are right, it was overriding the patrons commands from invenio-app-ils so I renamed it.

cds_ils/config.py
```diff
@@ -639,3 +641,5 @@ def _parse_env_bool(var_name, default=None):
 CDS_ILS_IMPORTER_PROVIDERS_ALLOWED_TO_DELETE_RECORDS = ["ebl", "safari"]
 
 RECORD_LEGACY_PID_TYPE = "lrecid"
+
+CDS_ILS_INDEX_LOCAL_ACCOUNTS = True
```
ntarocco, line 645: Can you please add a card to the board or push a commit to the deployed envs to avoid to forget to change this var?

ntarocco (approved):

PR #265, opened 2021-01-11
menu: Fixes responsive menu in mobile devices

jrcastro2:
## Before
image

## After
image

### DIFF
--- a/ui/src/overridden/components/Menu/Logo.js
+++ b/ui/src/overridden/components/Menu/Logo.js
@@ -14,7 +14,7 @@ export const Logo = ({ ...props }) => {
 export const LogoMobile = ({ ...props }) => {
   return (
     <h3>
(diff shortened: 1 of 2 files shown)


ntarocco (approved):

ui/src/overridden/components/Menu/Logo.js
```diff
@@ -14,7 +14,7 @@ export const Logo = ({ ...props }) => {
 export const LogoMobile = ({ ...props }) => {
   return (
     <h3>
-      <Link to="/" className="logo">
+      <Link to="/" className="logo size">
```
kprzerwa, line 17: I think class name size is used in SUI, I would suggest to use something like `.media-font-size`

kprzerwa (approved):

PR #274, opened 2021-01-18
migration: fix import document requests

kprzerwa:
fixed problem:
Screenshot 2021-01-15 at 15 04 27

the cancel reasons were quite random, while the ones we have are controlled
I moved the library notes to internal note field(https://github.com/inveniosoftware/invenio-app-ils/pull/1013), but the question is do we actually need it ?

### DIFF
--- a/cds_ils/migrator/document_requests/api.py
+++ b/cds_ils/migrator/document_requests/api.py
@@ -32,8 +32,8 @@ def migrate_document_request(record):
 
     if record["status"] == "proposal-put aside":
         state = "REJECTED"
-        reject_reason = record["library_notes"]
-        new_docreq.update(state=state, reject_reason=reject_reason)
+        new_docreq.update(state=state,
+                          reject_reason="OTHER")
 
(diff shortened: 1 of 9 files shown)


cds_ils/migrator/document_requests/api.py
```diff
@@ -32,8 +32,8 @@ def migrate_document_request(record):
 
     if record["status"] == "proposal-put aside":
         state = "REJECTED"
-        reject_reason = record["library_notes"]
-        new_docreq.update(state=state, reject_reason=reject_reason)
+        new_docreq.update(state=state,
+                          reject_reason="MIGRATED_UNKNOWN")
```
ntarocco, line 36: This is was called `MIGRATED` only if not mistaken in the other PR, but I suggested to use something else, e.g. `OTHER`.

jrcastro2 (approved):

KonstantinaStoikou (approved):

ntarocco (approved):

PR #278, opened 2021-01-20
migration: fix relation matching

kprzerwa:
* add update record to the migration script

* blocked by the previous PR


jrcastro2 (approved):

cds_ils/migrator/relations/api.py
```diff
@@ -88,11 +90,15 @@ def migrate_siblings_relation():
                 relation["relation_type"]
             )
 
+            if relation_type == OTHER_RELATION.name:
+                extra_metadata.update(
```
KonstantinaStoikou, line 94: Are you using the `extra_metadata` somewhere?

KonstantinaStoikou (approved):

cds_ils/migrator/relations/api.py
kprzerwa: 👍 nice ! good catch

PR #286, opened 2021-01-25
migration: migrate journal urls

### LINKED ISSUE #281: journals: migration rules for storage location
`866__b` -> `series.note` to store the physical location of the issues of the journal

KonstantinaStoikou:
Closes CERNDocumentServer/cds-dojson#264
Closes #281

Migrated fields: 85641, 85642, 8654_, 866__

### DIFF
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -339,9 +339,9 @@ def related_records(self, key, value):
             relation_description = description
             relation_type_tag = clean_val("x", value, str)
             if relation_type_tag:
-                if relation_type_tag.lower() == 'edition':
+                if relation_type_tag.lower() == "edition":
                     relation_type = EDITION_RELATION.name
-                elif relation_type_tag.lower() == 'language':
+                elif relation_type_tag.lower() == "language":
                     relation_type = LANGUAGE_RELATION.name
 
         if key == "787__" and "i" in value:
@@ -350,11 +350,11 @@ def related_records(self, key, value):
             {
                 "related_recid": clean_val("w", value, str, req=True),
                 "relation_type": relation_type,
-                "relation_description": relation_description
+                "relation_description": relation_description,
             }
         )
         _migration.update({"related": _related, "has_related": True})
-        raise IgnoreKey('_migration')
+        raise IgnoreKey("_migration")
     except ManualImportRequired as e:
         if key == "775__":
             e.subfield = "b or c"
@@ -856,7 +856,7 @@ def note(self, key, value):
     # merge all found notes
     _note = self.get("note", "")
     if _note:
-        _note = "{0} / {1}".format(_note, clean_val("a", value, str, req=True))
+        _note = "{0} \n{1}".format(_note, clean_val("a", value, str, req=True))
     else:
         _note = clean_val("a", value, str, req=True)
 
--- a/cds_ils/importer/providers/cds/rules/journal.py
+++ b/cds_ils/importer/providers/cds/rules/journal.py
@@ -85,11 +85,16 @@ def internal_notes(self, key, value):
     return {"value": clean_val("a", value, str, req=True)}
 
 
-@model.over("note", "^935__")
+@model.over("note", "(^935__)|(^866__)")
 @out_strip
 def note(self, key, value):
     """Translates note field."""
-    return clean_val("a", value, str, req=True)
+    notes_list = [self.get("note", "")]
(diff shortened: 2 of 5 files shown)


jrcastro2 (approved):

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -406,47 +406,56 @@ def open_access(self, key, value):
...
-    else:
-        # if none of the above, it is just external url
-        # attached to the document
-        return url
+    if key == "8654_":
+        url_dict = {
+            "value": clean_val("u", value, str),
+            "description": clean_val("y", value, str),
+        }
+        return url_dict
```
kprzerwa, line 419: in this case I would split those rules, since you return anyway, and we will not have additional level of if statements.

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -85,11 +85,19 @@ def internal_notes(self, key, value):
     return {"value": clean_val("a", value, str, req=True)}
 
 
-@model.over("note", "^935__")
+@model.over("note", "(^935__)|(^866__)")
 @out_strip
 def note(self, key, value):
     """Translates note field."""
-    return clean_val("a", value, str, req=True)
+    arr = [self.get("note", "")]
```
kprzerwa, line 92: naming police 🚓 
maybe something more descriptive for the list name, like
`list_of_notes`, ` list_of_notes_to_concatenate`

tests/migrator/test_books.py
```diff
@@ -490,7 +490,7 @@ def test_urls(app):
         )
         check_transformation(
             """
-            <datafield tag="8564" ind1=" " ind2=" ">
+            <datafield tag="856" ind1="4" ind2=" ">
```
kprzerwa, line 493: did you check for occurrences in the data of `856__` to be sure we can replace the ind1 for this test?

tests/migrator/test_journals.py
```diff
@@ -172,9 +170,13 @@ def test_note(app):
...
                 </subfield>
             </datafield>
+            <datafield tag="866" ind1=" " ind2=" ">
+                <subfield code="a">note1</subfield>
+                <subfield code="b">note2</subfield>
+            </datafield>
             """,
             {
-                "note": "The reports are catalogued individually and retrievable in the catalogue.",
+                "note": "The reports are catalogued individually and retrievable in the catalogue. / note1 / note2",
```
kprzerwa, line 179: 👍 
we chose to go with this solution, would you have any suggestions @ntarocco how to solve it  better?

ntarocco: Both `935` and `866` they are mapped to the `note` field, right? 
I am wondering if we add a `\n` instead: if in the form this is a `textarea`, new lines are `\n` I guess... But to be tested both when editing and when visualizing in the details page.

tests/migrator/test_books.py
KonstantinaStoikou: I didn't replace it, before it was `tag="8564"` and I moved the 4 to the ind1.

tests/migrator/test_journals.py
KonstantinaStoikou: In the textarea the `\n` works correctly but in the notes tab in the details page it shows all notes in the same line. It probably needs a javascript function to replace the `\n` with `<br />`

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -85,11 +85,19 @@ def internal_notes(self, key, value):
...
 
 
-@model.over("note", "^935__")
+@model.over("note", "(^935__)|(^866__)")
 @out_strip
 def note(self, key, value):
     """Translates note field."""
-    return clean_val("a", value, str, req=True)
+    notes_list = [self.get("note", "")]
+    if key == "935__":
```
kprzerwa, line 93: you could decrease the amount of code by removing this if statement (and the one below for checking if it is 866), since you are reading `a` subfield in each case anyway

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -85,11 +85,19 @@ def internal_notes(self, key, value):
...
+        notes_list.append(clean_val("a", value, str, req=True))
```
kprzerwa, line 94: why is the subfield a required?

KonstantinaStoikou: The code was like this before. I removed it now since in the schema it is not required.

PR #289, opened 2021-01-26
migration: fix relation note

kprzerwa:


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -336,13 +336,14 @@ def related_records(self, key, value):
...
         if key == "775__" and "b" in value:
             description = clean_val("b", value, str)
-            relation_type_tag = clean_val("x", value, str, default="")
-            if relation_type_tag.lower() == 'edition':
-                relation_type = EDITION_RELATION.name
-            elif relation_type_tag.lower() == 'language':
-                relation_type = LANGUAGE_RELATION.name
-            else:
-                relation_description = description
+            relation_description = description
```
KonstantinaStoikou, line 339: why not just do `relation_desciption = clean_val("b", value, str)`?

KonstantinaStoikou (approved):

jrcastro2 (approved):

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: because in case the field is empty, it is `None` and below I will try to call `.lower()` on None object

PR #290, opened 2021-01-26
migration: add relation when journal recid in publication info

### LINKED ISSUE #283: document: publication info - journal matching
Publication_info refers to two types of publications on CDS

A) record associated with journal which our library is subscribed to
B) NOT A)

in case A) the publication_info should be resolved as a series relation
in case B) publication_info stays as it was provided by cds-dojson

case A you can recognise via journal_record_legacy_recid field of migration

TASKS:
- [x] - map XML to JSON
`773__g` -> migration.journal_record_legacy_recid of the matching journal (has to be already migrated) 
- [x] create a relation of type serial between journal and the document (document should be of type periodical issue)
- [x] add `773__0` to ignored fields
- [x] clean the field `publication_info` of the records matched with a journal (also via script) as a redundant info)
- [x] review the rest of 773 fields (if implemented correctly, according to 
https://docs.google.com/spreadsheets/d/1OHMDiMfBD1Aam-kteIuBL9tEPK-C-Jevw-NcmGGxRvY/edit#gid=1257315908

KonstantinaStoikou:
Closes #283 
Closes https://github.com/orgs/inveniosoftware/projects/19#card-53138088

### DIFF
--- a/cds_ils/importer/providers/cds/ignore_fields.py
+++ b/cds_ils/importer/providers/cds/ignore_fields.py
@@ -68,6 +68,7 @@ CDS_IGNORE_FIELDS = {
     "695__a",
     "700__9",
     "710__5",
+    "773__0",  # on library request
     "773__r",  # publication_info/parent_report_number
     "773__w",  # inspire cnum (duplicated field with 035__9)
     "773__z",  # publication_info/parent_isbn
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -245,20 +245,29 @@ def publication_info(self, key, value):
     if x and o subfields are present simultaneously
     it concatenates the text
     """
+    _migration = self["_migration"]
     _publication_info = self.get("publication_info", [])
     for v in force_list(value):
         temp_info = {}
         pages = clean_pages_range("c", v)
         if pages:
             temp_info.update(pages)
+        volume = clean_val("v", v, str)
         temp_info.update(
             {
                 "journal_issue": clean_val("n", v, str),
                 "journal_title": clean_val("p", v, str),
-                "journal_volume": clean_val("v", v, str),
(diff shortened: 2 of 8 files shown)


cds_ils/importer/providers/cds/ignore_fields.py
```diff
@@ -68,6 +68,7 @@
     "695__a",
     "700__9",
     "710__5",
+    "773__0",
```
kprzerwa, line 71: could you add a comment ` # on library request ` ?

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -245,28 +245,38 @@ def publication_info(self, key, value):
...
                 "year": clean_val("y", v, int),
             }
         )
 
+        rel_recid = clean_val("g", value, str)
+        if rel_recid:
+            recids = _migration.get("journal_record_legacy_recids", [])
+            recids.append({"recid": rel_recid, "volume": volume})
+            _migration["has_journal"] = True
+            self["document_type"] = "PERIODICAL_ISSUE"
```
kprzerwa, line 270: 👍

cds_ils/migrator/relations/api.py
```diff
@@ -200,6 +201,23 @@ def link_records_and_serial(record_cls, search):
                     serial, record, SERIAL_RELATION, volume
                 )
 
+    def link_record_and_journal(record_cls, search):
```
kprzerwa, line 204: let's add test for the linking?

KonstantinaStoikou: I added them [here](https://github.com/CERNDocumentServer/cds-ils/blob/ac9f41bd3b0b24c000bf091d24e949ef5677b91d/tests/migrator/test_relations.py)

kprzerwa (approved):

PR #292, opened 2021-01-27
migration: Implement access type mapping

jrcastro2:
* closes https://github.com/CERNDocumentServer/cds-dojson/issues/263


cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -120,11 +121,15 @@ def children_records(self, key, value):
...
+        access_type_mapped = []
+        access_type_array = list(map(int, str(clean_val("x", value, str))))
+        for i in access_type_array:
+            access_type_mapped.append(mapping(ACCESS_TYPE, str(i)))
         _electronic_items.append(
             {
                 "subscription": clean_val("3", value, str),
                 "url": clean_val("u", value, str),
-                "access_type": clean_val("x", value, str),
+                "access_type": access_type_mapped,
```
kprzerwa, line 132: should be `access_restriction` the name of the field. like in the series schema

tests/migrator/test_journals.py
```diff
@@ -252,7 +252,39 @@ def test_migration(app):
                         {
                             "subscription": "v 1 (1992) -",
                             "url": "https://www.radioeng.cz/search.htm",
-                            "access_type": "6",
+                            "access_type": ["OPEN_ACCESS"],
+                            "note": "tada",
```
kprzerwa, line 256: 🎉  :D

tests/migrator/test_journals.py
```diff
@@ -252,7 +252,39 @@ def test_migration(app):
...
+                    "is_multipart": False,
+                    "has_related": False,
+                    "related": [],
+                    "record_type": "journal",
+                    "volumes": [],
+                    "electronic_items": [
+                        {
+                            "subscription": "v 1 (1992) -",
+                            "url": "https://www.radioeng.cz/search.htm",
+                            "access_type": ["RESTRICTED_PERPETUAL_ACCESS_BACKFILES",
```
kprzerwa, line 285: should be a list, did you find example of records which have more than one number in `85641x ` ?

kprzerwa: I checked, and it happens indeed

kprzerwa (approved):
LGTM!

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -128,9 +128,12 @@ def tags(self, key, value):
 def special_serials(self, key, value):
     """Translates serial fields."""
     _migration = self["_migration"]
-    _serials = _migration["serials"]
+    _serials = _migration.get("serials", False)
```
kprzerwa, line 131: `serials` is an empty list by default not `False`

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -124,11 +124,14 @@ def children_records(self, key, value):
 def access_urls(self, key, value):
     """Translates access urls field."""
     _access_urls = self.get("access_urls", [])
-
+    access_type_mapped = []
+    access_type_array = list(map(int, str(clean_val("x", value, str))))
```
kprzerwa, line 128: naming police 🚓 
access_type_array -> access_type_list

also, for this line:
```
str(clean_val("x", value, str)
```
clean_val function already returns str type (or None if empty) there is no need to cast it to string twice

tests/migrator/test_journals.py
```diff
@@ -232,12 +232,12 @@ def test_access_urls(app):
...
                 "access_urls": [
                     {
-                        "access_restriction": "8",
+                        "access_restriction": ['RESTRICTED_PACKAGE_DEAL'],
                         "description": "some text",
                         "value": "https://url.cern.ch/journal",
                     },
                     {
-                        "access_restriction": "4",
+                        "access_restriction": [None],
```
kprzerwa, line 240: is it correct, the None?
I think we might miss a value from the mapping if it is taken from a real example (please check on CDS)
if it does not occur on CDS, then we should raise UnexpectedValue if something missing from the mapping, in order to catch all the unexpected behaviours

cds_ils/importer/providers/cds/rules/base.py
jrcastro2: This is on purpose, this is used by journals too and they don't have `serials` in the `_migration`, so if this is the case we set it to false and in the if statemente below it will translate the `document_type`.

tests/migrator/test_journals.py
jrcastro2: It does not occur on CDS, I will change it to raise UnexpectedValue, thanks!

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: in the if statement below will work the same for the empty list, as follows:
```
if not result_a or not _serials:
```
while in the line 138, there is usage of `_serials` which assumes it is a list, and it will fail if `_serials` is False

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -128,9 +128,12 @@ def tags(self, key, value):
...
     for v in force_list(value):
         result_a = mapping(SERIAL, clean_val("a", v, str))
+        if not result_a or _serials is False:
```
kprzerwa, line 134: when assigning default value _serials  as empty list we most likely don't need ` or _serials is False` check

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -124,11 +125,17 @@ def children_records(self, key, value):
 def access_urls(self, key, value):
     """Translates access urls field."""
     _access_urls = self.get("access_urls", [])
-
+    access_type_mapped = []
+    access_type_list = list(map(int, clean_val("x", value, str)))
+    for i in access_type_list:
+        access_type = mapping(ACCESS_TYPE, str(i))
```
kprzerwa, line 131: `mapping` has `raise_exception` parameter which you could use instead of this `if` statement

kprzerwa (approved):

PR #293, opened 2021-01-27
languages: change standard to ISO 639-3

### LINKED ISSUE #287: languages: migration and vocabularies
change the languages standard to ISO 639-3
replace it in the migration rules and in the vocabularies values

KonstantinaStoikou:
Closes #287 

Removed book rules for cds provider and languages rule from journal because they were already in `base.py`


cds_ils/importer/providers/cds/rules/book.py
```diff
@@ -6,73 +6,3 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 """CDS-ILS MARCXML book rules."""
-
```
kprzerwa, line 9: why are these functions removed?

tests/data/documents.json
```diff
@@ -1,20 +1,20 @@
...
     "$schema": "https://127.0.0.1:5000/schemas/documents/document-v1.0.0.json",
-    "created_by": {"type": "script", "value":  "test"},
+    "created_by": { "type": "script", "value": "test" },
     "pid": "docid-1",
     "title": "The Gulf: The Making of An American Sea",
     "document_type": "BOOK",
     "authors": [{ "full_name": "Jack E. Davis" }],
     "abstracts": [{ "value": "This is an abstract" }],
-    "language": ["en"],
+    "language": ["eng"],
```
kprzerwa, line 10: shouldn't be uppercase?

tests/importer/test_safari_dojson.py
```diff
@@ -66,7 +66,7 @@ def test_safari_transformation(app):
                     {"scheme": "ISBN", "value": "9780814415467"},
                     {"scheme": "ISBN", "value": "9780814415474"},
                 ],
-                "languages": ["en"],
+                "languages": ["eng"],
```
kprzerwa, line 69: shouldn't be uppercase?

cds_ils/importer/providers/cds/rules/book.py
KonstantinaStoikou: Because they already existed in the `base.py` and they were not overriding them. Same for languages function in journal rules.

kprzerwa: you can remove the title and number of pages, but there is some kind of confusion with alternative_titles rule, please leave it as it is, I will take a look on how to solve it

KonstantinaStoikou: what about languages from the journal rules?

kprzerwa: this rule is the same, so you can remove it indeed. nice catch for the repeated code!

kprzerwa (approved):

PR #302, opened 2021-02-02
migration: fix multiple file sources for items

### LINKED ISSUE #296: migration: number of migrated items suspiciously low
investigate on the cause, it might be wrongly handled validation error

kprzerwa:
* closes #296

### DIFF
--- a/cds_ils/migrator/items/api.py
+++ b/cds_ils/migrator/items/api.py
@@ -75,10 +75,18 @@ def set_document_pid(record):
 
 def import_items_from_json(dump_file, rectype="item"):
     """Load items from json file."""
-    dump_file = dump_file[0]
     model, provider = model_provider_by_rectype(rectype)
-
     with click.progressbar(json.load(dump_file)) as bar:
+        error_logger.error(
+            "ITEMS: PROCESSING {0}".format(
+                dump_file
+            )
+        )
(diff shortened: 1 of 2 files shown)


cds_ils/migrator/items/api.py
```diff
@@ -75,10 +75,18 @@ def set_document_pid(record):
 
 def import_items_from_json(dump_file, rectype="item"):
     """Load items from json file."""
-    dump_file = dump_file[0]
     model, provider = model_provider_by_rectype(rectype)
-
     with click.progressbar(json.load(dump_file)) as bar:
+        error_logger.error(
```
jrcastro2, line 80: minor: Should it be info instead of error?
```suggestion
        error_logger.info(
```
question: Could you quickly explain when should we use each logger?

jrcastro2 (approved):

KonstantinaStoikou (approved):

cds_ils/migrator/items/api.py
kprzerwa: I use error because somewhere we have hardcoded logger level lower than info and logs with .info do not appear. I will test `LOGGING_CONSOLE_LEVEL = 'INFO'` and if it works I will change these accordingly

PR #311, opened 2021-02-04
edh: endpoint

jrcastro2:
* Updates EDH endpoint

image

### DIFF
--- a/cds_ils/patrons/serializers.py
+++ b/cds_ils/patrons/serializers.py
@@ -15,8 +15,14 @@ def serialize_on_loan_literature_info(loan):
     """Serialize loan information."""
     Document = current_app_ils.document_record_cls
     document = Document.get_record_by_pid(loan["document_pid"])
+    if "barcode" in loan["item"]:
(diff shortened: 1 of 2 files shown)


cds_ils/patrons/serializers.py
```diff
@@ -15,8 +15,14 @@ def serialize_on_loan_literature_info(loan):
     """Serialize loan information."""
     Document = current_app_ils.document_record_cls
     document = Document.get_record_by_pid(loan["document_pid"])
+    if "barcode" in loan["item"]:
```
ntarocco, line 18: This is ok totally ok! If you want to do it in a more pythonic way:
```
barcode = loan["item"].get("barcode", "")
```

ntarocco (approved):

cds_ils/patrons/serializers.py
jrcastro2: Tried that, but was getting this error:
`AttributeError: 'AttrDict' object has no attribute 'get'`

kprzerwa (approved):

PR #316, opened 2021-02-10
release: v1.0.0a31

kprzerwa:


docker-services.yml
```diff
@@ -29,7 +29,7 @@ services:
       - "15672:15672"
       - "5672:5672"
   es:
-    image: docker.elastic.co/elasticsearch/elasticsearch-oss:7.9.1
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:7.1.1
```
KonstantinaStoikou, line 32: why did you downgrade elasticsearch?

KonstantinaStoikou (approved):

docker-services.yml
kprzerwa: to have all the environments with the same version of elasticsearch, which caused that we missed the fact the boosting query will not work on sandbox

Issue #226, opened 2020-11-10
Vocabularies: Review institutions and studies fields


jrcastro2:
Currently institutions and studies are not vocabularies, even if institutions has a confirmed value <`CERN`> (see [here](https://codimd.web.cern.ch/uR0k-4yiTzW0BSnp7mDw0Q?view#Institutions)). After the clean up of the migrated data we need to verifiy if it's needed to create vocabularies for these fields.

EDIT: it will not be a vocabulary, check if it is migrated properly after the data cleaning

PR #318, opened 2021-02-10
extension: document extension fields

jrcastro2:
* Add vocabulary mapping for applicability and
  experiments fields
* Updates standard_review:checkdate to a string
* closes https://github.com/CERNDocumentServer/cds-ils/issues/226


kprzerwa (commented):
we wait for the fields numbers to be provided by the library

cds_ils/config.py
```diff
@@ -457,7 +457,7 @@ def _parse_env_bool(var_name, default=None):
             "marshmallow": SanitizedUnicode(required=True),
         },
         "standard_review:checkdate": {
-            "elasticsearch": "date",
```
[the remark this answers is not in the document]
jrcastro2: I didn't check all records but some of them doesn't have a date in the field 925__z. For example, this [record](https://cds.cern.ch/record/2690790/export/xm?ln=en) it's value is `Reviewed in December 2019`

cds_ils/vocabularies/data/document_experiments.json
```diff
@@ -524,11 +524,6 @@
 		"key": "HYBRID_PHOTODETECTOR",
 		"text": "HYBRID PHOTODETECTOR"
 	},
-	{
```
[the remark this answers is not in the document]
jrcastro2: This vocabularies were copied from [here](https://cds.cern.ch/kb/export?kbname=Experiments) but while adding the mapping for the migration I noticed that these values seem duplicated (they are very similar to the previous or equal), that's why I removed them.

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -407,9 +412,14 @@ def accelerator_experiments(self, key, value):
...
     sub_a = clean_val("a", value, str)
-    sub_e = clean_val("e", value, str)
+    sub_e = mapping(
+            EXPERIMENTS,
+            clean_val("e", value, str),
+            raise_exception=True,
+        )
     sub_p = clean_val("p", value, str)
 
+    # TODO: Add mapping for accelerators (Vocabulary)
```
ntarocco, line 422: Is this still valid? If yes, is there an issue for this? Please prefer issues with a tag BLOCKER instead, we will forget all TODOs otherwise

cds_ils/importer/providers/cds/rules/values_mapping.py
```diff
@@ -40,6 +40,1273 @@
     "migration": ["migration"],
 }
 
+ACQUISITION_METHOD = {
+    # possible user types (created_by.type/created_by.value)
+    "user": ["H", "R"],
```
ntarocco, line 45: what is the meaning of these letters? Can we have something more readable? Or are this values coming from the migrated data?

cds_ils/importer/providers/cds/rules/base.py
jrcastro2: Yes, the librarians are cleaning the values, and when they finish we can add the mapping. From a previous comment from @kprzerwa  I understand that we will wait until the librarians clean the values to merge this PR, so we can add the mapping in this same PR.

cds_ils/importer/providers/cds/rules/values_mapping.py
jrcastro2: Your comment is still valid, maybe @kprzerwa can answer better. I just want to clarify that this is duplicated code that I added wrongly and it's not part of this PR, I already removed it, sorry...

cds_ils/config.py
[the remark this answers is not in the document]
kprzerwa: then it should raise an error when we import, because the data is a wrong type. We shouldn't change our new data model because of the inconsistency in the old system

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: yes, could you please ask them when we could expect the values ?

ntarocco:
Some conflicts, could you please rebase?

cds_ils/vocabularies/data/document_insitutions.json
```diff
@@ -0,0 +1,42 @@
+[
```
ntarocco, line 1: typo in the filename `insitutions`

ntarocco (commented):
Did you check if you can edit custom fields in the document form editor?

jrcastro2:
> Did you check if you can edit custom fields in the document form editor?

Yes, we can edit except for the known issue with th `Study` field, it expects a list but it's a text field (this was also happening with `institution` but afet this change now it is a vocabulary field)
image

ntarocco (approved):

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -406,13 +425,36 @@ def accelerator_experiments(self, key, value):
...
+
+    val_a = clean_val("a", value, str)
+    val_e = clean_val("e", value, str)
+
+    def check_for_institution_in(val, value_to_check_for):
+        if value_to_check_for in val:
+            if value_to_check_for not in institutions:
+                institutions.append(value_to_check_for.upper())
+            val = val.replace(value_to_check_for, "")
+        return val.replace(" ", "")
```
kprzerwa, line 441: what if we have at some point an institution with two word name ? it will probably not happen for the migration but I wanted to make a point here that it would be safer to use `strip()` since in this way we replace all the spaces

tests/migrator/test_books.py
```diff
@@ -1042,13 +1042,13 @@ def test_extensions(app):
             <datafield tag="925" ind1=" " ind2=" ">
                 <subfield code="i">applicable at CERN</subfield>
                 <subfield code="p">Expert ICS-25.160</subfield>
-                <subfield code="z">Reviewed in December 2019</subfield>
+                <subfield code="z">20040621</subfield>
```
kprzerwa, line 1045: are there any examples of records from legacy with this date format ?
the tests were written to mirror the real records, f.e. https://cds.cern.ch/record/2690790
I am asking because in this case we have to somehow normalise the data

tests/migrator/test_books.py
```diff
@@ -1059,48 +1059,79 @@ def test_extensions(app):
...
                 <subfield code="p">Expert ICS-25.160</subfield>
                 <subfield code="v">withdrawn</subfield>
-                <subfield code="z">Reviewed in December 2019</subfield>
+                <subfield code="z">20040621</subfield>
             </datafield>
             """,
             {
                 "extensions": {
-                    "unit_accelerator": ["CERN LHC"],
+                    "unit_accelerator": ["CERN_SPS"],
```
kprzerwa, line 1114: shouldn't be only `SPS`?

jrcastro2: I kept it as `CERN SPS` because we got the values from the [accelerators KB](https://cds.cern.ch/kb/export?kbname=Accelerators).

tests/migrator/test_books.py
jrcastro2: Your are right, I checked the records and all of them have `Reviewed in XXX`. I will normalise it, thanks!

tests/migrator/test_books.py
kprzerwa: yes I see but the goal was to separate the accelerators/experiments from institution, so it is an inconsistency with the rest

jrcastro2: Ok, I will remove `CERN` from the accelerators and leave only `SPS`. Thanks!

kprzerwa (approved):

PR #319, opened 2021-02-11
migration: multiple values in created_by field

### LINKED ISSUE #303: 916__w: multiple subfields
in many records, in the field `916__` we have multiple occurrences of the subfield `w`. (it is the creation date in format YYYYWW, where WW is the week number).

For this case, the migration rule should pick up the **oldest** date from all the occurrences as the creation date and ignore the other

kprzerwa:
* closes #303

### DIFF
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -72,7 +72,18 @@ def created(self, key, value):
                 }
             )
             self["created_by"] = _created_by
-            date = clean_val("w", value, int, regex_format=r"\d{6}$")
+
+            # used to compare dates, to find the oldest publication date
+            # when multiple dates present in the field
+            date = 999999
+            date_values = clean_val("w", value, int, regex_format=r"\d{6}$",
+                                    multiple_values=True)
+            if type(date_values) is list:
+                for x in date_values:
+                    if x < date:
+                        date = x
+            else:
+                date = date_values
             if date:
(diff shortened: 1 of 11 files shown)


jrcastro2 (approved):

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -72,7 +72,16 @@ def created(self, key, value):
                 }
             )
             self["created_by"] = _created_by
-            date = clean_val("w", value, int, regex_format=r"\d{6}$")
+
+            date = 99999
```
ntarocco, line 76: date `99999`? What kind of date is that???? :)

KonstantinaStoikou (approved):

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: :) this is a dirty hack, I didn't have a better idea there. The problem is that in this field we have date in format `YYYYWW` where `WW` is week number, and there might be two subfields `w`. We are supposed to choose the oldest date, so the initial 999999 is for comparing reasons. I thought it is not worth it to spend time for more sophisticated solutions.

ntarocco: Ok, makes sense, I would just add a comment (even another PR) to explain this... if you read this code in one year I would wonder what happened here... :)

ntarocco (approved):

ntarocco:
It requires rebase and fix conflicts

Issue #308, opened 2021-02-03
journals: 980__ 690C and 960 field


kprzerwa:
- [x] Unexpected value: "PERI" or "31" problem
write a migration rule for journals, where
if 980__ or 690C__ == PERI:  ignore field
if 960 == 31 ignore field
else : raise exception for Unexpected value

- [x] check the low rate of importing journals, if there is no other problems

jrcastro2:
Currently we are importing 77% of journals (1808/2343).

Most frequent errors:
- MIGRATION RULE MISSING {'555__a'}
- MIGRATION RULE MISSING {'85642x'}
- MIGRATION RULE MISSING {'866__z'}
- [UNEXPECTED INPUT VALUE]  MARC FIELD: *697C_*, input value: GroupableOrderedDict((('a', 'virTH')))
  - Different unexpected input values 
- MIGRATION RULE MISSING {'938__i'}
- [MISSING REQUIRED FIELD]  MARC FIELD: *210__*, input value: GroupableOrderedDict((('9', 'J. Phys., B')))

Less frequent errors:
- MIGRATION RULE MISSING {'866__c'}
- Impossible to convert to JSON 'eitems_ebl'
-  [MISSING REQUIRED FIELD]  MARC FIELD: *780__*,
-  ERROR in xml_to_json_dump: MIGRATION RULE MISSING {'222__b'}
-  MIGRATION RULE MISSING {'85641a'}
-  [UNEXPECTED INPUT VALUE]  MARC FIELD: *041__*, input value: GroupableOrderedDict((('a', 'jap'))),
-  [MISSING REQUIRED FIELD]  MARC FIELD: *780__*,
-  MIGRATION RULE MISSING {'246_3C'} 
-  [MISSING REQUIRED FIELD]  MARC FIELD: *022__*,

PR #339, opened 2021-02-16
migration: journal fixes

jrcastro2:
* Adds ignore rule for document type
* Fixes agency_code bug, it was preventing the records indexing
* Fixes _migration bug, there where some missing fields who caused 'NoneType' error
* closes #https://github.com/CERNDocumentServer/cds-ils/issues/308

ignored rules requested in https://cernbox.cern.ch/index.php/apps/onlyoffice/edit/__myprojects/digital-repositories/Services_Projects/CDS/CDS-ILS/CDS%20-%20Library%20shared/missing-rules-values-journals.xlsx

### DIFF
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -45,7 +45,10 @@ def recid(self, key, value):
 @model.over("agency_code", "^003")
 def agency_code(self, key, value):
     """Control number identifier."""
-    return value
+    if isinstance(value, str):
+        return value
+    else:
+        raise IgnoreKey("agency_code")
 
 
 @model.over("created_by", "^859__")
--- a/scripts/bootstrap
+++ b/scripts/bootstrap
@@ -11,7 +11,7 @@ set -e
 script_path=$(dirname "$0")
 
 # Install application code and entrypoints from 'setup.py'
-pip install --use-deprecated=legacy-resolver -e $script_path/..[all]
+pip install -e $script_path/..[all]
 # Build assets
 invenio collect -v
 invenio webpack buildall
--- a/cds_ils/importer/providers/cds/cds.py
+++ b/cds_ils/importer/providers/cds/cds.py
@@ -12,10 +12,10 @@ from cds_ils.importer.base_model import Base
 from cds_ils.importer.base_model import model as model_base
 
 
-def get_helper_dict():
+def get_helper_dict(record_type):
     """Return migration extra data."""
     _helper_dict = dict(
-        record_type="document",
+        record_type=record_type,
         volumes=[],
         volumes_identifiers=[],
         volumes_urls=[],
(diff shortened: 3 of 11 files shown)


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -157,9 +160,12 @@ def doc_type_mapping(val):
             return mapping(DOCUMENT_TYPE, val)
 
     for v in force_list(value):
-        val_a = doc_type_mapping(clean_val("a", v, str))
+        clean_val_a = clean_val("a", v, str)
+        val_a = doc_type_mapping(clean_val_a)
         val_b = doc_type_mapping(clean_val("b", v, str))
-
+        if ((key == "980__" or key == "690C_") and clean_val_a == "PERI") \
```
kprzerwa, line 166: in this case it should be a separate rule, moved to journals.py, because it is specific for journals, especially in terms of the values. 
For example in the data set of documents (not journals) we will have `PERI` value in the subfield `a`. From the point of view of the book it should raise UnexpectedValue. From the point of view of the journal this value is correct.

scripts/bootstrap
```diff
@@ -11,7 +11,7 @@ set -e
 script_path=$(dirname "$0")
 
 # Install application code and entrypoints from 'setup.py'
-pip install --use-deprecated=legacy-resolver -e $script_path/..[all]
+pip install -e $script_path/..[all]
```
kprzerwa, line 14: maybe install before this line the explicit pip version, like we do in the ./run-tests.sh, otherwise if someone has newest pip, it will hang trying to install dependencies (unless you tested the newest and there is no problem)

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -45,7 +45,10 @@ def recid(self, key, value):
 @model.over("agency_code", "^003")
 def agency_code(self, key, value):
     """Control number identifier."""
-    return value
+    if isinstance(value, str):
```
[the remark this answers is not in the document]
jrcastro2: I added this check because some records do have a `datafield` like this `<datafield tag=\"003\" ind1=\" \" ind2=\" \">\n  </datafield>\n` instead of a `controlfield` like `<controlfield tag=\"003\">SzGeCERN</controlfield>\n`.
With the `controlfield` we get the value as a `String` but with the `datafields` we get an empty `GroupableOrderedDict`, this was causing that the record could not be correctly indexed in ES since the dynamic mapping recognized it as a `text` but with the empty `GroupableOrderedDict`, it was trying to index an empty array, something like:
```
agency_code:[
       __order__: [ ]
]
```

cds_ils/importer/providers/cds/rules/base.py
jrcastro2: I see, thanks! :+1:

scripts/bootstrap
jrcastro2: I just tested it with the latest pip (`pip 21.0.1`) and there are no problems

kprzerwa (approved):

PR #340, opened 2021-02-17
migration: mint item,acq order, borrowing req pids

### LINKED ISSUE #297: migration: acq orders: not the full data set imported
one dump chunk file has 1000 records. In some cases only a fraction (950) of the records of the 1000 is imported. 
Investigate why.

KonstantinaStoikou:
* mint item,acq order, borrowing req legacy pids
* allow updates in CDSRecordDumpLoader

Closes #297


cds_ils/migrator/json_record_loader.py
```diff
@@ -47,6 +52,20 @@ def create_record(
                     object_type="rec",
                     object_uuid=record_uuid,
                 )
+
+                config = current_app.config
```
kprzerwa, line 56: not a blocker: I would put these lines in a separate function (the mapping) in order to keep the `create_record` method's code cleaner

kprzerwa (approved):

jrcastro2 (approved):

PR #349, opened 2021-03-01
[BLOCKED by RIAILS#383] IE: default message

### LINKED ISSUE #24: Test responsive design and Internet Explorer
Verify that UI looks good on mobile and when using Internet Explorer

jrcastro2:
* Adds IE check and unsupported
  browser message
* closes #24
* needs https://github.com/inveniosoftware/react-invenio-app-ils/pull/383

Twitter and other apps that doesn't support IE get redirected to Edge automatically, in order to get this we need to register our app in the [IE compatibility list](https://docs.microsoft.com/en-us/microsoft-edge/web-platform/ie-to-microsoft-edge-redirection#request-an-update-to-the-ie-compatibility-list), by doing this Windows will automatically redirect the users to Edge.
Alternatively to be able to reproduce the behaviour to open Edge when accesing our site from IE we could add this code (but I recommend to register the app in the IE compatibility list so we don't need to manage/mantain this):
```
if(isIE) {
    window.location = 'microsoft-edge:' + window.location;
    setTimeout(function() {
      window.location = 'https://go.microsoft.com/fwlink/?linkid=2135547';
    }, 1);
  }
```
## Unsupported browser message 
image

P.S: Writting style of the message was inspired/copied from facebook, if we want to be more polite we can use `Unsupported browser`, instead of `Improve Your Experience`.


ui/public/js/check-for-IE.js
```diff
@@ -0,0 +1,79 @@
+window.onload = function() {
+    const isIE = !!document.documentMode;
```
jrcastro2, line 2: After some searches I think this is the most reliable way of detecting IE, check property [here](https://www.w3schools.com/jsref/prop_doc_documentmode.asp). There is nice analysis done by people from jquery explaining by documentMode is good enough, check it [here](https://github.com/jquery/jquery/issues/4386).

jrcastro2:
For the new browser icons, I searched for them in google with the filter `Comercial and other license`, so they don't have copyrights.

ui/public/js/check-for-IE.js
```diff
@@ -0,0 +1,78 @@
...
+    if (isIE) {
+      // Adds styling
+      var style = document.createElement('style');
+      style.innerHTML =
```
kprzerwa, line 6: could we just add a small css file to avoid writing it this way ?

ui/public/js/check-for-IE.js
```diff
@@ -0,0 +1,78 @@
...
+      imgChrome.src = '/images/chrome-logo.png';
+      aChrome.href = 'https://www.google.com/chrome/';
+      pChrome.innerHTML = 'Google Chrome';
+      aChrome.appendChild(imgChrome);
+      aChrome.appendChild(pChrome);
+
+      let imgFirefox = document.createElement('img');
+      let aFirefox = document.createElement('a');
+      let pFirefox = document.createElement('p');
+      imgFirefox.src = '/images/firefox-logo.png';
```
kprzerwa, line 56: I would probably link to their logo instead of adding is as a static for our website.

kprzerwa (commented):
thanks for taking care of this.
added some small comments below.
In general: shouldn't we solve it in react-invenio-app-ils this way ?

jrcastro2:
> thanks for taking care of this.
> added some small comments below.
> In general: shouldn't we solve it in react-invenio-app-ils this way ?

I am not sure this is an ILS responsibility but we can add the message this same way (if you all agree).

ui/src/index.js
```diff
@@ -1,18 +1,22 @@
...
-ReactDOM.render(
-  <Router history={history}>
-    <OverridableContext.Provider value={overriddenCmps}>
-      <InvenioILSApp config={config} />
-    </OverridableContext.Provider>
-  </Router>,
-  document.getElementById('app')
-);
+// Checks if browser is IE (unsupported)
+const isIE = !!document.documentMode;
```
ntarocco, line 11: Since the `index.js` is loaded anyway, I would suggest to put this in `react-invenio-app-ils` and export it to avoid duplications, otherwise everyone creating their own ILS will have to copy/paste this.
The idea is that here you will simply do:
```
import { IEFallback, history, InvenioILSApp } from '@inveniosoftware/react-invenio-app-ils';

...

if (!isIE) {
...
} else {
  IEFallback()
}
```

The IEFallback function contains all the JS  in the `check-for-IE` file. I would also add the CSS there, or even inline for this...
WDYT?

ui/public/index.html
```diff
@@ -26,6 +29,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>CERN Library Catalogue</title>
+    <script type="text/javascript" src="js/check-for-IE.js"></script>
```
ntarocco, line 32: No need, see comment below

ntarocco:
> ~For the new browser icons, I searched for them in google with the filter `Comercial and other license`, so they don't have copyrights.~
> 
> EDIT: Following @kprzerwa suggestion, logos are linked.

I added a comment about that, not sure what was her comment

ntarocco:
> > thanks for taking care of this.
> > added some small comments below.
> > In general: shouldn't we solve it in react-invenio-app-ils this way ?
> 
> I am not sure this is an ILS responsibility but we can add the message this same way (if you all agree).
> 
> EDIT: We agreed to add the message to ILS, this is the [PR](https://github.com/inveniosoftware/react-invenio-app-ils/pull/383).

I think it is indeed the responsibility of `react-invenio-app-ils` because without it cannot work :'(

ntarocco:
`if we want to be more polite we can use Unsupported browser, instead of Improve Your Experience.` I agree with you, we should use `Unsupported browser` because the app does not work at all....
For me `Improve Your Experience` means that it works, but degraded, which is not the case.

ui/public/js/check-for-IE.js
```diff
@@ -0,0 +1,55 @@
...
+    imagesDiv.style.cssText += 'display: inline-flex; padding: 0.5em';
+    h1.innerHTML = 'Improve Your Experience';
+    p.innerHTML =
+      "You're using a web browser we don't support.<br>" +
+      'Try one of these options to have a better experience on CERN Library.';
+
+    let imgChrome = document.createElement('img');
+    let aChrome = document.createElement('a');
+    let pChrome = document.createElement('p');
+    imgChrome.src = 'https://www.google.com/chrome/static/images/chrome-logo.svg';
```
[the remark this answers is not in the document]
jrcastro2: Couldn't find links to logos with full availability guarantee so I added the images to the public folder. The size of the images is 5.8kb, 2.8kb, and 2.8kb. Is this good enough?

ntarocco (approved):

PR #363, opened 2021-03-03
ES: replaces execute()

jrcastro2:
* Replaces execute() from ES querys capable of
  returning more than 10 hits with scan()
* closes https://github.com/inveniosoftware/invenio-app-ils/issues/1048

### DIFF
--- a/cds_ils/patrons/serializers.py
+++ b/cds_ils/patrons/serializers.py
@@ -47,15 +47,15 @@ def patron_loans_to_dict(patron_loans):
     :return: dict from patron's loan.
     :rtype: dict
     """
-    literatures_on_loan_results = patron_loans["active_loans"].hits.hits
+    literatures_on_loan_results = patron_loans["active_loans"].scan()
     literatures_on_loan = [
-        serialize_on_loan_literature_info(loan["_source"])
+        serialize_on_loan_literature_info(loan)
         for loan in literatures_on_loan_results
     ]
 
-    loan_requests_results = patron_loans["pending_loans"].hits.hits
+    loan_requests_results = patron_loans["pending_loans"].scan()
     loan_requests = [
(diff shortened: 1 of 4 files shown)


KonstantinaStoikou (approved):

cds_ils/patrons/serializers.py
```diff
@@ -47,15 +47,15 @@ def patron_loans_to_dict(patron_loans):
     :return: dict from patron's loan.
     :rtype: dict
     """
-    literatures_on_loan_results = patron_loans["active_loans"].hits.hits
+    literatures_on_loan_results = patron_loans["active_loans"].scan()
     literatures_on_loan = [
-        serialize_on_loan_literature_info(loan["_source"])
+        serialize_on_loan_literature_info(loan)
```
ntarocco, line 52: just wondering if you were able to validate this change since the field changes...

cds_ils/patrons/views.py
```diff
@@ -69,12 +69,11 @@ def get(self, patron, **kwargs):
         ]
         active_loans = search_by_patron_item_or_document(
             patron["id"], filter_states=active_loan_states
-        ).execute()
-
+        )
```
ntarocco, line 72: is this change correct? there is no execute/scan called.

jrcastro2: Yes, instead of calling it here we will call `scan` in the [serializer](https://github.com/CERNDocumentServer/cds-ils/pull/363/files#diff-c3b7becdc2b2b6fb688d80b45e1f753413d1222d51ef2dee168f3611541f281dR50)

cds_ils/patrons/serializers.py
jrcastro2: I tested it locally and it works correctly

Issue #361, opened 2021-03-03
journal: physical location field


kprzerwa:
- [x] add field location/description for the eitems and for the physical item to the series data model (schema), to describe the physical location of the items, and the description of the e-items
- [x] fix field 362 migration rule accordingly
- [x] display the field in the backoffice
- [x] display the field on the frontsite

The field (or fields) are created here to migrate the data:

for eitems of journal:
```
    start and end volume (field 362__a)
    year (362__a)
```
to describe which eitems they have available

for physical items of journal:
```
start-end volume (866__a)
year (866__a)
location (866__b)
```

to describe which physical items they have and where they are stored

examples:
https://cds.cern.ch/record/2737239?ln=en physical items
https://cds.cern.ch/record/2722082?ln=en electronic items

PR #370, opened 2021-03-08
[BLOCKED by IAILS#1059] journals: volumes rules

jrcastro2:
* Adds rules to translate the physical_volumes
  and electronic_volumes_description fields
* Needs https://github.com/inveniosoftware/invenio-app-ils/pull/1059
* closes https://github.com/CERNDocumentServer/cds-ils/issues/361

### DIFF
--- a/cds_ils/importer/providers/cds/rules/journal.py
+++ b/cds_ils/importer/providers/cds/rules/journal.py
@@ -89,14 +89,12 @@ def internal_notes(self, key, value):
     return {"value": clean_val("a", value, str, req=True)}
 
 
-@model.over("note", "(^935__)|(^866__)")
+@model.over("note", "(^935__)")
 @out_strip
 def note(self, key, value):
     """Translates note field."""
     notes_list = [self.get("note", "")]
     notes_list.append(clean_val("a", value, str))
-    if key == "866__":
-        notes_list.append(clean_val("b", value, str))
 
     return " \n".join(filter(None, notes_list))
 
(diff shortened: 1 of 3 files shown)


cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -108,20 +106,30 @@ def publisher(self, key, value):
...
-    _migration.update(
-        {
-            "electronic_items": _electronic_items,
-        }
-    )
+@model.over("electronic_volumes_description", "(^362__)")
+def electronic_volumes_description(self, key, value):
+    """Translates electronic volumes description field."""
+    description = clean_val("a", value, str, req=True)
+    if description.endswith('-'):
```
kprzerwa, line 113: I think this hyphen is there on purpose, could you check with the library ?

cds_ils/templates/invenio_app_ils_circulation/mail/cds_request.html
```diff
@@ -21,7 +21,7 @@
 
 your loan request for <a href="{{ spa_routes.HOST }}{{ spa_routes.PATHS['literature']|format(pid=document.pid) }}">"{{ document.full_title }}"</a> has been <b>received</b>.
 <br/><br/>
-You selected: <b>"{{ circulation_delivery_methods[loan.delivery.method] }}"</b>
+You selected: <b>{{ circulation_delivery_methods[loan.delivery.method] }}</b>
```
ntarocco, line 24: Thank you!

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -108,20 +106,30 @@ def publisher(self, key, value):
...
+        description = description[:-1]
+    return description
```
ntarocco, line 115: Does it make sense to also add a `strip`?
```suggestion
    return description.strip()
```
So that as in the test `v 1 (1992) -` becomes `v 1 (1992)`

ntarocco (approved):

cds_ils/importer/providers/cds/rules/journal.py
jrcastro2: You are right, it's on purpose and we should keep them, library answer:
"it is important to keep it, those fields are meant to describe a time period, when there is no 2nd volume/year it is because the collection or publication continues till now"

kprzerwa (approved):

PR #371, opened 2021-03-09
standard: rules

jrcastro2:
* Adds rule for 084__c
* Fixes relation between standard and
  volumes of multiparts
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/75

### DIFF
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -24,8 +24,8 @@ from cds_ils.importer.errors import ManualImportRequired, \
     MissingRequiredField, UnexpectedValue
 from cds_ils.importer.providers.cds.cds import model
 from cds_ils.importer.providers.cds.rules.utils import clean_email, \
-    clean_pages_range, clean_val, extract_volume_number, filter_list_values, \
-    get_week_start, out_strip, replace_in_result
+    clean_val, extract_volume_number, filter_list_values, get_week_start, \
+    out_strip, replace_in_result
 from cds_ils.importer.providers.cds.rules.values_mapping import ACCELERATORS, \
     ACQUISITION_METHOD, APPLICABILITY, ARXIV_CATEGORIES, COLLECTION, \
     DOCUMENT_TYPE, EXPERIMENTS, EXTERNAL_SYSTEM_IDENTIFIERS, \
@@ -298,9 +298,9 @@ def publication_info(self, key, value):
     _publication_info = self.get("publication_info", [])
     for v in force_list(value):
         temp_info = {}
-        pages = clean_pages_range("c", v)
+        pages = clean_val("c", v, str)
         if pages:
-            temp_info.update(pages)
+            temp_info.update(pages=pages)
         volume = clean_val("v", v, str)
         temp_info.update(
             {
@@ -387,22 +387,19 @@ def publication_additional(self, key, value):
     empty = not bool(_publication_info)
     for i, v in enumerate(force_list(value)):
         temp_info = {}
-        pages = clean_pages_range("k", v)
+        pages = clean_val("k", v, str)
         if pages:
-            temp_info.update(pages)
+            temp_info.update(pages=pages)
         rel_recid = clean_val("b", v, str)
         if rel_recid:
             _related.append(
                 {
                     "related_recid": rel_recid,
                     "relation_type": OTHER_RELATION.name,
-                    "relation_description": "chapter of",
+                    "relation_description": "is chapter of"
                 }
             )
             _migration.update({"related": _related, "has_related": True})
-        n_subfield = clean_val("n", v, str)
-        if n_subfield.upper() == "BOOK":
-            temp_info.update({"material": "BOOK"})
         if not empty and i < len(_publication_info):
             _publication_info[i].update(temp_info)
         else:
@@ -820,7 +817,14 @@ def languages(self, key, value):
 def subject_classification(self, key, value):
     """Translates subject classification field."""
     prev_subjects = self.get("subjects", [])
-    _subject_classification = {"value": clean_val("a", value, str, req=True)}
+    if key == "084__":
+        _subject_classification = {
+            "value": clean_val("c", value, str, req=True)
+        }
+    else:
+        _subject_classification = {
+            "value": clean_val("a", value, str, req=True)
+        }
     if key == "080__":
         _subject_classification.update({"scheme": "UDC"})
     elif key.startswith("082"):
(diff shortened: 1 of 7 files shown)


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -373,23 +373,21 @@ def publication_additional(self, key, value):
...
-            temp_info.update(pages)
+            temp_info.update(pages=pages)
         rel_recid = clean_val("b", v, str)
         if rel_recid:
             _related.append(
                 {
                     "related_recid": rel_recid,
                     "relation_type": OTHER_RELATION.name,
-                    "relation_description": "chapter of"
+                    "relation_description": "is chapter of the book"
```
ntarocco, line 385: Is it ok to put the word `book` and not literature? is it always a book the parent?

cds_ils/migrator/api.py
```diff
@@ -61,10 +61,11 @@ def import_documents_from_dump(sources, source_type, eager, include):
                         )
 
 
-def import_record(dump, model, pid_provider, legacy_id_key="legacy_recid"):
+def import_record(dump, model, pid_provider,
+                  legacy_id_key="legacy_recid", mint_id=None):
     """Import record in database."""
     record = CDSRecordDumpLoader.create(
-        dump, model, pid_provider, legacy_id_key
+        dump, model, pid_provider, legacy_id_key, mint_id
```
ntarocco, line 68: what is a `mint_id`?

cds_ils/migrator/relations/api.py
```diff
@@ -116,25 +116,30 @@ def migrate_document_siblings_relation():
...
-
-            # try to find sibling in series
-            if related_sibling is None:
-                try:
-                    related_sibling = get_record_by_legacy_recid(
-                        series_class,
-                        legacy_pid_type,
-                        relation["related_recid"],
+                continue
+            rec_type = related_sibling["$schema"].split("/")[-1].split("-")[0]
```
ntarocco, line 120: why not checking with `instance of` or `type`?

cds_ils/migrator/relations/api.py
```diff
@@ -116,25 +116,30 @@ def migrate_document_siblings_relation():
...
+            # If it's series it must be a multipart, so we search for the
+            # first volume and link it.
```
ntarocco, line 122: Maybe better to assert it than? What happens if it is not the case?
`assert series["MODE_OF_ISSUANCE"] == MULTI...`

cds_ils/migrator/relations/api.py
```diff
@@ -116,25 +116,30 @@ def migrate_document_siblings_relation():
...
+            if rec_type == "series":
+                doc_search = document_search.query(
```
ntarocco, line 124: I would suggest to use a `term` search instead?

cds_ils/migrator/relations/api.py
```diff
@@ -116,25 +116,30 @@ def migrate_document_siblings_relation():
...
+                    "query_string",
+                    query='relations.multipart_monograph.pid_value:"{}"'
+                                            .format(related_sibling["pid"])
+                )
+                doc_result = doc_search.execute()
+                if doc_result.hits.total.value > 0:
```
ntarocco, line 130: is it ok to have for example 2 of them? More than 1?

ntarocco (commented):
A few comments, not sure if they are very relevant! Can you rebase?

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: i think we could leave it as `is chapter of` to make it generic for all document types

cds_ils/importer/providers/cds/rules/multipart.py
```diff
@@ -234,24 +234,18 @@ def volumes_titles(self, key, value):
...
-                    raise UnexpectedValue(
-                        subfield="y", message=" unrecognized publication year"
-                    )
-            _insert_volume(_migration, volume_number, obj)
+        volume_number = extract_volume_number(
+            val_n, raise_exception=True, subfield="n"
+        )
+        obj = {"title": val_p or volume_title}
+        if val_y:
+            if re.match("\\d+", val_y) and 1600 <= int(val_y) <= 2021:
```
kprzerwa, line 242: what if we have an older book ?

cds_ils/importer/providers/cds/rules/multipart.py
```diff
@@ -234,24 +234,18 @@ def volumes_titles(self, key, value):
...
+                obj["publication_year"] = int(val_y)
```
kprzerwa, line 243: `publication_year` is a string

cds_ils/migrator/api.py
kprzerwa: same question

cds_ils/migrator/relations/api.py
kprzerwa: I am not sure what are you trying to solve here, let's maybe discuss in a call

cds_ils/importer/providers/cds/rules/base.py
jrcastro2: I changed it as requested in the [issue](https://github.com/CERNDocumentServer/cds-migrator-kit/issues/75). ( description: "is chapter of the book" )

cds_ils/importer/providers/cds/rules/multipart.py
jrcastro2: This code was already here, I only removed the restriction for the on digit volumes, I am not sure about this code, maybe we should remove the `1600 <= int(val_y) <= 2021:` constraint? WDYT?

cds_ils/migrator/relations/api.py
jrcastro2: Because it returns a dict, not a Document or Series instance

cds_ils/migrator/relations/api.py
jrcastro2: I am trying to relate the MULTIVOLUMES-X which have the legacy_recid from the series minted but not the legacy_recid from the documents, when relating them we need to search for the Multipart and return the first document of that multipart, to match the standards with that document, if I am not mistaken. But we can discuss it better in a call.

cds_ils/migrator/relations/api.py
jrcastro2: Yes, Anne said to match it with the first one created.

cds_ils/migrator/api.py
jrcastro2: It's used to pass the id we want to mint for that record, since we were not doing it. Already fixed in a better way by @kprzerwa in another PR, so I reverted these changes.

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: OK I think we should discuss this on the standup

kprzerwa (approved):

PR #376, opened 2021-03-11
migration: stop minting title as legacy_recid

### LINKED ISSUE #374: legacy recid minted incorrectly
Screenshot 2021-03-10 at 22 07 53

KonstantinaStoikou:
Closes #374

### DIFF
--- a/cds_ils/migrator/json_record_loader.py
+++ b/cds_ils/migrator/json_record_loader.py
@@ -65,7 +65,7 @@ class CDSRecordDumpLoader(object):
                 )
                 legacy_pid_type = cls.get_legacy_pid_type_by_provider(provider)
 
-                if legacy_pid_type:
+                if legacy_pid_type and legacy_id_key == "legacy_recid":
                     legacy_recid_minter(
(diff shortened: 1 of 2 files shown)


cds_ils/migrator/json_record_loader.py
```diff
@@ -65,7 +65,7 @@ def create_record(
                 )
                 legacy_pid_type = cls.get_legacy_pid_type_by_provider(provider)
 
-                if legacy_pid_type:
+                if legacy_pid_type and legacy_id_key == "legacy_recid":
```
kprzerwa, line 68: if we integrate this change we will not mint for these types anymore
```
CDS_ILS_BORROWING_REQ_LEGACY_PID_TYPE = "lillb"
CDS_ILS_ACQ_ORDER_LEGACY_PID_TYPE = "lacqid"
CDS_ILS_ITEM_LEGACY_PID_TYPE = "litmid"
```
also legacy recid type is missing for series which causes it to mint legacy pid of **document** type for **series** which causes further problems in migration

PR #379, opened 2021-03-12
migration: concat all series numbers when multiple

KonstantinaStoikou:


ntarocco:
LGTM, but I would prefer @kprzerwa to check it! :)

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -887,7 +887,7 @@ def clean_conference_info_fields(v, required=True):
                     "value": clean_val("g", v, str),
                 }
             ],
-            "series": str(series_number),
+            "series": str(series_number).strip("[]"),
```
kprzerwa, line 890: why do you use the `.strip("[]")` ?

KonstantinaStoikou: because the `series_number` is an array e.g. `[2, 3]` and I want it to be a string like `"2, 3"` because in the schema of conference_info.series is a string

kprzerwa (approved):

PR #387, opened 2021-03-15
document: rules

jrcastro2:
* Fix document rules
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/78
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/79

### DIFF
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -824,7 +824,7 @@ def subject_classification(self, key, value):
     if key == "080__":
         _subject_classification.update({"scheme": "UDC"})
     elif key.startswith("082"):
-        _subject_classification.update({"scheme": "Dewey"})
+        _subject_classification.update({"scheme": "DEWEY"})
     elif key == "084__":
         sub_2 = clean_val("2", value, str)
         if sub_2 and sub_2.upper() in SUBJECT_CLASSIFICATION_EXCEPTIONS:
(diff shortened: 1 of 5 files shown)


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -1045,7 +1045,8 @@ def table_of_content(self, key, value):
     ).strip()
     if text != "--":
         chapters = re.split(r"; | -- |--", text)
-        return chapters
+        chapters = [elem.strip(' ') for elem in chapters]
+        return list(filter(None, chapters))
```
ntarocco, line 1049: Can it be empty string too (not only `None`) and you have to filter it out?

ntarocco (approved):

kprzerwa (approved):

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -1045,7 +1045,8 @@ def table_of_content(self, key, value):
     ).strip()
     if text != "--":
         chapters = re.split(r"; | -- |--", text)
-        return chapters
+        chapters = [elem.strip(' ') for elem in chapters]
```
kprzerwa, line 1048: for this you can use the decorator function, if I remember correctly it was `@out_strip`, automatically cleans the list of empty values

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: for this you can use the decorator function, if I remember correctly it was `@out_strip`, automatically cleans the list of empty values

jrcastro2: No, with None we already filter the empty strings. With `None` we check if it's true or not for each element of the list and return the ones that are true, if an element is empty string it will be False and filtered out (that's why we apply the strip just above to remove any blank space on empty lines)

PR #388, opened 2021-03-15
[BLOCKED by RIAILS release] extension fields fixes

ntarocco:
* Fixes extensions fields definitions
* Removes required property for standards
* Makes `shelf` mandatory field in Item editor schema
* Change default provider type to a valid value


cds_ils/migrator/default_records.py
```diff
@@ -74,10 +74,8 @@ def create_unknown_provider():
         "pid": MIGRATION_PROVIDER_PID,
         "name": "Migrated Unknown Provider",
         "legacy_ids": ["0"],
-        "type": "UNKNOWN",
-        "notes": """This Provider is used whenever we had provider ID 0
-        in CDS Acquisition and ILL data.
-        """,
+        "type": "LIBRARY",
```
ntarocco, line 77: For reviewers: the `type` is a vocab, it should be a valid value, I choose the first between LIBRARY and VENDOR

kprzerwa (approved):

PR #392, opened 2021-03-16
ui: display standard number in an prominent place

KonstantinaStoikou:
Depends on https://github.com/inveniosoftware/react-invenio-app-ils/pull/410

Screenshot 2021-03-17 at 14 31 20

Multiple values that don't fit in the same line:
Screenshot 2021-03-17 at 15 43 19

Multiple values that fit in the same line:
Screenshot 2021-03-17 at 15 41 22

Mobile:
Screenshot 2021-03-17 at 14 31 37

Screenshot 2021-03-17 at 15 43 29

Screenshot 2021-03-17 at 15 41 38

### DIFF
--- a/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentPanel/StandardNumber.js
+++ b/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentPanel/StandardNumber.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import { Label } from 'semantic-ui-react';
+import PropTypes from 'prop-types';
+import _isEmpty from 'lodash/isEmpty';
+
+export const StandardNumber = ({ metadata, ...props }) => {
+  const renderLabels = numbers => {
+    return numbers.map(number => (
+      <Label
+        className="default-margin-bottom standard-number"
+        color="grey"
+        image
+        key={number}
+      >
+        Standard number
+        <Label.Detail>{number}</Label.Detail>
+      </Label>
+    ));
+  };
+
+  let standardNumbers = [];
+  if (metadata.identifiers) {
+    standardNumbers = metadata.identifiers.filter(
+      sn => sn.scheme === 'STANDARD_NUMBER'
+    );
+    standardNumbers = standardNumbers.map(a => a.value);
+  }
+
+  return _isEmpty(standardNumbers) ? null : (
+    <div>{renderLabels(standardNumbers)}</div>
+  );
+};
+
+StandardNumber.propTypes = {
+  metadata: PropTypes.object.isRequired,
+};
--- a/ui/src/semantic-ui/site/elements/label.overrides
+++ b/ui/src/semantic-ui/site/elements/label.overrides
@@ -1,3 +1,10 @@
 /*******************************
          Site Overrides
 *******************************/
+
+/* Frontsite */
+@{fs-parent-selector} {
+    .standard-number {
+        margin-left: 0;
+    }
+}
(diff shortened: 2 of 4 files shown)


ui/src/overridden/frontsite/Document/DocumentDetails/DocumentPanel/StandardNumber.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Label } from 'semantic-ui-react';
+import PropTypes from 'prop-types';
+
+export const StandardNumber = ({ metadata, ...props }) => {
+  if (metadata.identifiers) {
+    const standardNumbers = metadata.identifiers.filter(
+      sn => sn.scheme === 'STANDARD_NUMBER'
+    );
+    if (standardNumbers[0]) {
```
ntarocco, line 10: it might be an empty list:
```suggestion
    if (standardNumbers) {
```

You can also do:
```
return standardNumbers ? <> : null;
```

ui/src/semantic-ui/site/elements/label.overrides
```diff
@@ -1,3 +1,15 @@
...
          Site Overrides
 *******************************/
+
+/* Frontsite */
+@{fs-parent-selector} {
+    .standard-number {
+        margin-left: 0;
+
+        .detail {
+            margin: -.5833em .5em -.5833em -.533em !important;
```
ntarocco, line 11: Are these margins ok??? :) Looks like a weird number to me!

KonstantinaStoikou: it's because in the [semantic ui label](https://react.semantic-ui.com/elements/label/) the half label with the darker background is in the opposite side so I had to use the same margins with the label but in the opposite order.

ntarocco: I see now, thank you for the explanation! Eventually I have other proposals then to avoid this:

Screenshot 2021-03-17 at 11 43 30

```
<Label as='div' labelPosition='right'>
      <Label>
        Standard number
      </Label>
      <Label as='span' basic>
        ISO-21312-2131233-31212
      </Label>
    </Label>
```

Using a button but on hover it changes, maybe we can find a solution to that:

Screenshot 2021-03-17 at 11 45 34


```
<Button as='div' labelPosition='right'>
      <Button icon>
        Standard number
      </Button>
      <Label as='span' basic>
        ISO-21312-2131233-31212
      </Label>
    </Button>
```

ntarocco (approved):

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentPanel/StandardNumber.js
```diff
@@ -0,0 +1,25 @@
...
+    standardNumbers = metadata.identifiers.filter(
+      sn => sn.scheme === 'STANDARD_NUMBER'
+    );
+    standardNumbers = standardNumbers.map(a => a.value);
+  }
+
+  return _isEmpty(standardNumbers) ? null : (
+    <Label className="default-margin-bottom standard-number" color="grey" image>
+      Standard number
+      <Label.Detail>{standardNumbers.join(', ')}</Label.Detail>
```
kprzerwa, line 18: I don't know if you discussed it earlier but I would prepare to have a separate label for each number, not to concatenate it in one label. @ntarocco @jrcastro2 WDYT ?

jrcastro2: I agree, I think it would be more visual

kprzerwa (approved):

PR #393, opened 2021-03-17
migration: files

jrcastro2:
* Adds format when importing files
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/82

### DIFF
--- a/cds_ils/migrator/eitems/api.py
+++ b/cds_ils/migrator/eitems/api.py
@@ -169,9 +169,7 @@ def process_files_from_legacy():
                 )
 
                 # get filename
-                file_name = file_dump["description"]
-                if not file_name:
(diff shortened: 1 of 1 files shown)


cds_ils/migrator/eitems/api.py
```diff
@@ -169,8 +169,10 @@ def process_files_from_legacy():
                 )
 
                 # get filename
-                file_name = file_dump["description"]
-                if not file_name:
+                if "description" in file_dump:
```
ntarocco, line 172: Am I a bit worried that there is no sanitization at all of this before setting it as filename.
It is not very important because in Invenio it is just a string in a db table, but what if it is super long?
Do you have any example in CDS?

kprzerwa (approved):

PR #397, opened 2021-03-17
migration: fix journal access url

kprzerwa:


cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -133,22 +131,26 @@ def access_urls(self, key, value):
...
     val_x = clean_val("x", value, str)
     access_type_list = list(map(int, val_x if val_x else []))
+
     for i in access_type_list:
         access_type = mapping(ACCESS_TYPE, str(i), raise_exception=True)
         access_type_mapped.append(access_type)
+
+    sub_3 = clean_val("3", value, str)
+    sub_z = clean_val("z", value, str)
+    electronic_volumes_description = f"{sub_3} ({sub_z})"
```
ntarocco, line 141: super minor
```suggestion
    electronic_volumes_description = f"{sub_3} ({sub_z})".trim()
```
or strip()

so if one is empty, there is no extra space

cds_ils/importer/providers/cds/rules/journal.py
```diff
@@ -175,30 +177,30 @@ def related_records(self, key, value):
     _migration = self.get("_migration", {})
     _related = _migration.get("related", [])
     description = None
+    sequence_order = None
     relation_type = OTHER_RELATION.name
-
+    relation_type_tag = clean_val("x", value, str)
+    if relation_type_tag:
```
ntarocco, line 183: what is the relation_type is none of the one we support?
Don't you want to have an:
```
else:
    raise
```
?

cds_ils/migrator/series/xml_series_loader.py
```diff
@@ -42,28 +44,68 @@ def create(cls, dump, rectype):
...
+            return series
+        except PIDAlreadyExists as e:
+            allow_updates = current_app.config.get(
+                "CDS_ILS_MIGRATION_ALLOW_UPDATES"
+            )
+            if not allow_updates:
+                raise e
+            # update document if already exists with legacy_recid
+            legacy_pid_type = current_app.config[
+                "CDS_ILS_SERIES_LEGACY_PID_TYPE"
```
ntarocco, line 90: Not for now, but all these `CDS_ILS_SERIES_LEGACY_PID_TYPE` should have been **hardcoded** values, not config. They should never change.

PR #401, opened 2021-03-18
fix migration rules for license

ntarocco:


cds_ils/importer/providers/springer/rules/document.py
```diff
@@ -295,5 +295,5 @@ def id_isbns(self, key, value):
 def eitem_internal_note(self, key, value):
     """Translate item internal note."""
     _eitem = self.get("_eitem", {})
-    _eitem["internal_note"] = clean_val("a", value, str)
+    _eitem["internal_notes"] = clean_val("a", value, str)
```
ntarocco, line 298: @kprzerwa I am not 100% sure if this correct, can you check?

kprzerwa: it is correct, nice catch

cds_ils/migrator/xml_to_json_dump.py
```diff
@@ -46,16 +46,10 @@ def created(self):
...
-        if self.latest_only:
-            self.revisions.append(
-                self._prepare_revision(self.data["record"][0])
-            )
-        else:
-            self.revisions = self._prepare_revision(self.data["record"])
+        """Prepare revisions."""
+        # get only the latest in any case
+        latest = self.data["record"][-1]
+        self.revisions = [self._prepare_revision(latest)]
```
kprzerwa, line 52: thanks for fixing it

kprzerwa (approved):

PR #402, opened 2021-03-18
migration: validate vocabulary values

KonstantinaStoikou:
Closes https://github.com/orgs/inveniosoftware/projects/19#card-57011896


kprzerwa (commented):
I would prefer to verify it on the level of the rules, not the whole record, because it adds complexity with the mapping @ntarocco WDYT?

cds_ils/cli.py
```diff
@@ -229,6 +229,8 @@ def run_command(command, catch_exceptions=False, verbose=True):
     run_command("vocabulary index opendefinition opendefinition --force")
     # languages
     run_command("vocabulary index languages --force")
+    # countries
+    run_command("vocabulary index languages --force")
```
KonstantinaStoikou, line 233: why do we need to run the same command twice?

ntarocco: Because it should be `countries` instead! Thank you for spotting this, fixed! :) 👍🏼

cds_ils/migrator/cli.py
```diff
@@ -65,6 +68,7 @@ def migration():
         "acq-order": records_formatter,
         "eitem": eitems_formatter,
         "relation": records_formatter,
+        "vocabularie": records_formatter,
```
kprzerwa, line 71: uff sorry about the `s`

kprzerwa (approved):
LGTM!
If you don't mind I will check it out before merging and I will try to run the migration on it to make sure that the migration scripts will run correctly

jrcastro2 (approved):

PR #408, opened 2021-03-19
migration: files

jrcastro2:
* Adds exceptions for files migrations
* closes https://github.com/CERNDocumentServer/cds-migrator-kit/issues/92

### DIFF
--- a/cds_ils/importer/providers/cds/rules/base.py
+++ b/cds_ils/importer/providers/cds/rules/base.py
@@ -157,7 +157,10 @@ def special_serials(self, key, value):
     _migration = self["_migration"]
     _serials = _migration.get("serials", [])
     for v in force_list(value):
-        result_a = mapping(SERIAL, clean_val("a", v, str))
+        value_a = clean_val("a", v, str)
+        if value_a and value_a == "YELLOW REPORT":
+            _migration["is_yellow_report"] = True
+        result_a = mapping(SERIAL, value_a)
         if result_a:
             _serials.append(
                 {
--- a/cds_ils/migrator/eitems/api.py
+++ b/cds_ils/migrator/eitems/api.py
@@ -145,7 +145,6 @@ def process_files_from_legacy():
 
         try:
             for file_dump in document["_migration"]["files"]:
-
                 # check if url migrated from MARC
                 url_in_marc = [
                     item
--- a/cds_ils/importer/providers/cds/cds.py
+++ b/cds_ils/importer/providers/cds/cds.py
@@ -22,6 +22,7 @@ def get_helper_dict(record_type):
         serials=[],
         has_serial=False,
         is_multipart=False,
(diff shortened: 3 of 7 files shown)


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -177,14 +177,21 @@ def special_serials(self, key, value):
...
     def doc_type_mapping(val):
         if val:
             return mapping(DOCUMENT_TYPE, val)
 
     for v in force_list(value):
-        val_a = doc_type_mapping(clean_val("a", v, str))
-        val_b = doc_type_mapping(clean_val("b", v, str))
+        value_a = clean_val("a", v, str)
+        value_b = clean_val("b", v, str)
+        if key == "690C_" and (
```
kprzerwa, line 189: if I remember correctly, if 690C will be equal YELLOW REPORT then in line 193 it will fail the migration anyway, since yellow report is not in the mapping, please correct me if I am wrong

cds_ils/migrator/eitems/api.py
```diff
@@ -145,7 +145,11 @@ def process_files_from_legacy():
 
         try:
             for file_dump in document["_migration"]["files"]:
-
+                if (
+                    file_dump.get("format") == ".ps.gz"
```
kprzerwa, line 149: why this format ?

kprzerwa (approved):

cds_ils/importer/providers/cds/rules/base.py
jrcastro2: It will throw `ERROR: [UNEXPECTED INPUT VALUE]` and the record is migrated anyway, at least locally (this is one[ record example](https://cds.cern.ch/record/923393))

cds_ils/migrator/eitems/api.py
jrcastro2: Requested in the [issue](https://github.com/CERNDocumentServer/cds-migrator-kit/issues/92):
```
The files with .ps.gz shouldn't be imported at all (in principle, there is the corresponding version in PDF)
```

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: yes, but from the code you wrote it means that value it is expected, so it is confusing. If you take a look on the two previous functions, you are already processing 690C_ there. Keeping 3 different functions for one MARC field is a little confusing. I can explain in detail in a call

cds_ils/migrator/eitems/api.py
kprzerwa: now when I am thinking, it would be just easier not to add the files to ["_migration"]["files"] if the file is .ps.gz or if is it a yellow report, instead of adding them and just skipping in the iteration

kprzerwa (approved):

PR #412, opened 2021-03-22
migration: open_access

jrcastro2:
* Migrates open acces field for eitems
* https://github.com/CERNDocumentServer/cds-migrator-kit/issues/84

rules \ links | EBL | Safari | Local files |  rest of links(e-items)
--------------- | ----- | --------- | --------- | --------
 "Open Access" in 8564_y | Restricted (by default) | Restricted (by default) | not taken into account(files restriction) | Not restricted
"Open Access" in 0247_q | Restricted (by default) | Restricted (by default) | not taken into account(files restriction)  | Not restricted
"Open Access" in 536__r* | Restricted (by default) | Restricted (by default) | not taken into account(files restriction)  | Not restricted

* **536__r***: This one is globally it means that applies to all e-items (except for EBL, Safari and local files) present in the record, while 8564_y and 0247_q only apply to one concrete e-item.


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -510,38 +509,48 @@ def urls(self, key, value):
...
+    elif sub_u.startswith("https://learning.oreilly.com/library/view/"):
+        eitem["open_access"] = False
+        eitems_safari.append(eitem)
+        self["_migration"]["eitems_has_safari"] = True
     # local files
     elif all(
         [elem in sub_u for elem in ["cds", ".cern.ch/record/", "/files"]]
     ):
-        eitems_files.append(url)
+        del eitem["open_access"]
```
kprzerwa, line 544: could you add a comment for this line? That the open_access is controlled by files restrictions itself?

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -510,38 +509,48 @@ def urls(self, key, value):
     sub_u = clean_val("u", value, str, req=True)
 
     eitems_ebl = self["_migration"]["eitems_ebl"]
+    eitems_safari = self["_migration"]["eitems_safari"]
     eitems_external = self["_migration"]["eitems_external"]
     eitems_proxy = self["_migration"]["eitems_proxy"]
     eitems_files = self["_migration"]["eitems_file_links"]
 
-    url = {"value": sub_u}
+    eitem = {"url": {"value": sub_u}, "open_access": False}
```
kprzerwa, line 517: I wouldn't name it eitem, because it is not this record type yet, it is just a dict representing the eitem's url. It looks confusing when you return eitem["url"] in the end.
From what I have understood from the rest of the code, adding the open_access here is not necessary, since you do it based on the provider (EBL/SAFARI/FILE/PROXY/OTHERS) and you also rely on the `['_migration']['eitems_open_access']` to solve the problem for eitems_external

cds_ils/migrator/eitems/api.py
```diff
@@ -344,3 +354,35 @@ def migrate_ebl_links(raise_exceptions=True):
...
+    click.echo("Found {} documents with safari links.".format(search.count()))
+
+    for hit in search.scan():
+        # make sure the document is in DB not only ES
+        document = document_class.get_record_by_pid(hit.pid)
+        click.echo("Processing document {}...".format(document["pid"]))
+
+        try:
+            for item in document["_migration"]["eitems_safari"]:
+                eitem = create_eitem(document["pid"], open_access=False)
```
kprzerwa, line 373: if you set here `open_access= False` why did you add it as a property in the urls in the migration rules?

cds_ils/migrator/eitems/api.py
```diff
@@ -236,12 +241,17 @@ def migrate_ezproxy_links(raise_exceptions=True):
         Document = current_app_ils.document_record_cls
         document = Document.get_record_by_pid(hit.pid)
         click.echo("Processing document {}...".format(document["pid"]))
-
-        for url in document["_migration"]["eitems_proxy"]:
+        open_access = document["_migration"]["eitems_open_access"]
```
kprzerwa, line 244: why did you add open_access depending on this field when login_required=True ?

cds_ils/migrator/eitems/api.py
jrcastro2: I thought it would be more consistent if all the urls had the `open_access` property even if it's not used, but you are right, it doesn't make sense for Safari and EBL since they are always restricted. I will remove the property, thanks.

cds_ils/migrator/eitems/api.py
jrcastro2: I am not sure I understand the question. Do you mean that if login_required=True then we should not take into account open_access?

kprzerwa: could you check with @ntarocco about how login_required influences the open_access? maybe we have it documented somewhere, he worked quite heavily on that part

ntarocco: Here more details: https://github.com/CERNDocumentServer/cds-ils/issues/162

jrcastro2: It seems that there is no relation between login_required and open_access fields, so we respect marcXML fields values.

cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -510,38 +509,54 @@ def urls(self, key, value):
...
 
     eitems_ebl = self["_migration"]["eitems_ebl"]
+    eitems_safari = self["_migration"]["eitems_safari"]
     eitems_external = self["_migration"]["eitems_external"]
     eitems_proxy = self["_migration"]["eitems_proxy"]
     eitems_files = self["_migration"]["eitems_file_links"]
 
-    url = {"value": sub_u}
+    def translate_open_access(item, field):
+        if sub_y:
```
kprzerwa, line 518: variable `sub_y` is taken from outer scope of this function, and then you pass it as a `field` parameter, like this, it is very easy to get a bug. 
We should also add a small comment on what is sub_y and what is sub_u, above, to be understandable when reading the code

tests/migrator/xml_imports/test_migrate_proceedings.py
```diff
@@ -72,37 +72,29 @@ def test_migrate_record(datadir, base_app):
                 'Karen Yagdjian: Integral transform approach to time-dependent partial differential equations.'],
             'abstract': 'This book collects lectures given by the plenary speakers at the 10th International ISAAC Congress, held in Macau, China in 2015. The contributions, authored by eminent specialists, present some of the most exciting recent developments in mathematical analysis, probability theory, and related applications. Topics include: partial differential equations in mathematical physics, Fourier analysis, probability and Brownian motion, numerical analysis, and reproducing kernels. The volume also presents a lecture on the visual exploration of complex functions using the domain coloring technique. Thanks to the accessible style used, readers only need a basic command of calculus.',
             '_migration': {
+                **get_helper_dict(record_type="document"),
```
kprzerwa, line 75: about this, I have left it as a raw dict for a reason here, in order to not be able to change internal migration fields so easily as only to change them in the helper dict. Its because then many things depend on this dictionary and I would us to be reminded about that by failing tests

kprzerwa (approved):

PR #426, opened 2021-03-30
acq/ill: decode utf-8 characters in notes

KonstantinaStoikou:
Closes https://github.com/orgs/inveniosoftware/projects/19#card-57920310


cds_ils/migrator/utils.py
```diff
@@ -153,6 +153,16 @@ def get_acq_ill_notes(record):
     item_info = record.get("item_info")
     if item_info:
         result = "item info: {0}\n\n".format(item_info.strip())
+        try:
+            result = (
+                result.encode("ascii")
+                .decode("unicode-escape")
+                .encode("iso-8859-1")
```
kprzerwa, line 160: why do you use so many encode-decode ? especially why `iso-8859-1`  ?

kprzerwa:
test for the encoding missing

cds_ils/migrator/utils.py
KonstantinaStoikou: See [here](https://stackoverflow.com/questions/49756071/python-3-6-utf-8-to-unicode-conversion-string-with-double-backslashes/49756591)

kprzerwa (approved):

PR #432, opened 2021-04-07
[BLOCKED]migration: migrate login_required for series

KonstantinaStoikou:
Depends on https://github.com/inveniosoftware/invenio-app-ils/pull/1077

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -37,6 +37,7 @@ from invenio_app_ils.patrons.api import PATRON_PID_TYPE
 from invenio_app_ils.permissions import authenticated_user_permission, \
     backoffice_permission, loan_extend_circulation_permission, \
     patron_owner_permission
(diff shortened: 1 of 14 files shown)


cds_ils/series/serializers/custom_fields.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Series custom serializer functions."""
+
+from flask import current_app
+
+
+def format_login_required_urls(metadata):
```
ntarocco, line 13: I think this is duplicated right? Maybe move this code somewhere to re-use it without duplication?
You can pass as param the URLs list of the entire metadata eventually.

ntarocco (approved):

PR #433, opened 2021-04-07
migration: fix author roles

kprzerwa:
* fix document types and tags
* fix experiments
* add material to identifiers vocabulary
* bugfixes


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -1212,22 +1204,26 @@ def medium(self, key, value):
...
+            val_a.upper().replace("-", ""),
+            raise_exception=True,
+        )
 
     for barcode in barcodes:
         current_item = {
             "barcode": barcode,
-            "medium": _medium,
         }
+        if _medium:
```
ntarocco, line 1225: I don't know if my comment makes sense, but `medium` is mandatory in the schema if I am not wrong

ntarocco (approved):

cds_ils/importer/providers/cds/rules/base.py
kprzerwa: we have the default medium on the import time. The rule is just trying to map any of the allowed mediums. Now I see that it would make sense to move the default here but I am worried to change it at this point. WDYT ?

PR #435, opened 2021-04-08
piwik: track homepage

KonstantinaStoikou:


ui/public/index.html
```diff
@@ -29,6 +29,39 @@
...
+          pkBaseURL +
+          "piwik.js' type='text/javascript'%3E%3C/script%3E"
+      )
+    );
+  </script>
+  <script type="text/javascript">
+    try {
+      var piwikTracker = Piwik.getTracker(
+        pkBaseURL + 'piwik.php',
+        process.env.REACT_APP_INVENIO_PIWIK_ID
```
ntarocco, line 50: you can test this locally, you can change the `.env` [here](https://github.com/CERNDocumentServer/cds-ils/blob/master/ui/.env) and it should be working.

KonstantinaStoikou: I tested it and it seems to be working (I didn't get any errors and the variable is set correctly)

ui/public/index.html
```diff
@@ -29,6 +29,39 @@
...
+        : 'http://piwik.web.cern.ch/';
+    document.write(
+      unescape(
+        "%3Cscript src='" +
+          pkBaseURL +
+          "piwik.js' type='text/javascript'%3E%3C/script%3E"
+      )
+    );
+  </script>
+  <script type="text/javascript">
```
ntarocco, line 46: You can remove this double closing/opening of JS.
Can you also add a `if PROD`? We need to find the right variable... otherwise locally it will keep sending requests...

kprzerwa (approved):

PR #517, opened 2021-05-21
ldap: add patrons mailbox

kprzerwa:
* refactoring

### DIFF
--- a/cds_ils/ldap/errors.py
+++ b/cds_ils/ldap/errors.py
@@ -1,0 +1,20 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019-2020 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""CDS-ILS ldap exceptions."""
+import sys
+
+
+class InvalidLdapUser(Exception):
+    """Invalid user exception."""
+
+    def __init__(self, *args, log_func=None):
+        """Constructor."""
+        super().__init__(*args)
+        print(args[0], file=sys.stderr)
+        if log_func:
+            log_func("ldap_user_has_no_email")
--- a/cds_ils/ldap/serializers.py
+++ b/cds_ils/ldap/serializers.py
@@ -1,0 +1,126 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019-2020 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""CDS-ILS ldap serializers."""
+from functools import partial
+
+from invenio_accounts.models import User
+from invenio_oauthclient.models import UserIdentity
+from invenio_userprofiles import UserProfile
+
+from cds_ils.ldap.errors import InvalidLdapUser
+
+
+def serialize_ldap_user(ldap_user_data, log_func=None):
+    """Create ldap user."""
+
+    def serialize(ldap_user_data):
+        decoded_data = {}
+        for key, value in ldap_user_data.items():
+            # get first value of each field from LDAP
+            # decode bytes to UTF-8
+            decoded_data[key] = value[0].decode("utf8")
+        serialized_data = dict(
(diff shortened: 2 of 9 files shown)


cds_ils/ldap/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+        """Constructor."""
+        self.is_valid_user = True
+        self.is_duplicate = False
+        self.employee_id = ldap_user_data["employeeID"][0].decode("utf8")
+        self.data = {}
+        if self.validate_required(ldap_user_data):
+            self.data = self.serialize(ldap_user_data)
+            self.validate_user_data(log_func)
+
+    def serialize(self, ldap_user_data):
```
ntarocco, line 30: I have the feeling that all these methods should be private?

cds_ils/ldap/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+
+class LDAPUser:
+    """LDAPUser serializer class."""
+
+    def __init__(self, ldap_user_data, log_func=lambda x: x):
+        """Constructor."""
+        self.is_valid_user = True
+        self.is_duplicate = False
+        self.employee_id = ldap_user_data["employeeID"][0].decode("utf8")
+        self.data = {}
```
ntarocco, line 25: Maybe I am wrong, the diff is large, but it looks to me that `data` is required not to be empty, when used.
Isn't it better to `raise` if `validate_required` return `False`?

cds_ils/ldap/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+
+"""CDS-ILS ldap serializers."""
+import sys
+
+from invenio_accounts.models import User
+from invenio_oauthclient.models import UserIdentity
+from invenio_userprofiles import UserProfile
+
+
+class LDAPUser:
```
ntarocco, line 17: Just in general: to me this does look more a way to force the OOP :) LDAPUser looks like more to a data structure, and not to an obj with behavior. It might be better to create a `factory` function instead.
The "contract/interface" of this object says that you can call the various methods, which instead it looks like they are only needed when creating the obj and should not be really used otherwise.

```
create_user(...):
    return LDAPUser(...)
```
and is should not be possible to create/modify the obj (immutable).

ntarocco (approved):

cds_ils/ldap/serializers.py
kprzerwa: I was considering it while implementing, I will fix it since you think the same :)

cds_ils/ldap/serializers.py
kprzerwa: yes, the class felt unnatural to me as well, but it was my go-to pattern. I will ask for more details on the factory

cds_ils/ldap/errors.py
```diff
@@ -0,0 +1,20 @@
...
+import sys
+
+
+class InvalidLdapUser(Exception):
+    """Invalid user exception."""
+
+    def __init__(self, *args, log_func=None):
+        """Constructor."""
+        super().__init__(*args)
+        print(args[0], file=sys.stderr)
```
ntarocco, line 18: is `print` correct here? I saw it is used also above.

ui/src/overridden/backoffice/PatronMetadata/PatronMetadata.js
```diff
@@ -5,15 +5,29 @@ import {
...
   const rightTable = [
     { name: 'Person ID', value: patron.metadata.person_id },
     { name: 'Department', value: patron.metadata.department },
+    {
+      name: 'CERN Profile',
+      value: patron.metadata.person_id ? (
+        <a
+          target="_blank"
+          rel="noreferrer"
+          href={config.PATRONS.phonebookURLPrefix + patron.metadata.person_id}
```
ntarocco, line 25: 👍🏼

ntarocco (approved):
LGTM!
Next time, if you remember, it would be nice to see a little screenshot to see how it looks. :)

cds_ils/ldap/errors.py
kprzerwa: it was used before, when the update/import is triggered from cli

PR #524, opened 2021-05-28
ldap: save remote account changes

kprzerwa:


cds_ils/ldap/serializers.py
```diff
@@ -124,3 +125,4 @@ def update(self, ldap_user):
         ra.extra_data["mailbox"] = ldap_user["remote_account_mailbox"]
         self.user.email = ldap_user["user_email"]
         self.user_profile.full_name = ldap_user["user_profile_full_name"]
+        db.session.commit()
```
ntarocco, line 128: Ideally it is better to keep only one `db.session.commit()` at the end, in case things need to be rollbacked.
I guess you needed this for the indexing of patrons? If yes, why don't you change the code [here](https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/ldap/api.py#L156) and save in a list which patrons should be indexed and do it after the last db commit()?

kprzerwa: because it is commits to RemoteAccount table, then later here https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/ldap/api.py#L168 we access the table again (https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/patrons/api.py#L27), but the new data is not stored

PR #545, opened 2021-06-09
importer: preview of import 

### LINKED ISSUE #321: importer: preview of changes
Implement a preview of the changes which are about to be done via importer.

The preview should be visible after uploading an XML file, but before performing any changes in the DB. The Librarian should be able to choose if they want to continue the import based on the preview.

In the minimal implementation it could be json preview, similar to what was done for cds-migrator-kit.
At the moment, in CDS, librarians can see the converted/mapped XMLs that are about to be imported in a textarea and can check each value.

## Tasks

- [ ] Implement the preview: this should be the first mandatory step. The outcome could be simply the JSON output that will be imported:
```
r = Record(dict(...data...))
json.dump(r)
```
The extra complexity is when creating e.g. eitem/item which require an imported document, a created PID. Can we fake such PID and put e.g. `document_pid:<newly imported PID>`?
- [ ] After the preview, the button `Import` will appear
- [ ] Change the table of results after import to be able to support the display of more of 1K results for a single job. The easiest is to simply show a very simple table, probably with pagination or infinite scrolling (how does the webpage react to 1K elements?)

### LINKED ISSUE #322: importer: add recid of the record to the report
in the importer report there is no indication of which record failed on the import time (besides the order in the file matching the order in the report). 
In the report add recid (from field 001 of marcxml, mapped as provider_recid) to the report list entry (especially for the failed records, where the only information is 'No document imported')

kprzerwa:
* refactoring
 * closes #321, closes #322 
Screenshot 2021-06-11 at 17 42 40
Screenshot 2021-06-11 at 17 43 45
Screenshot 2021-06-11 at 17 43 04
Screenshot 2021-06-11 at 17 42 55


cds_ils/importer/api.py
```diff
@@ -43,61 +42,68 @@ def process_dump(data, provider, mode, source_type):
...
+def validate_import(provider, mode, source_type):
+    """Validate import."""
     source_type = source_type or "marcxml"
     assert source_type in ["marcxml"]
 
+    # check if the record is in delete mode
     if provider not in current_app.config[
         "CDS_ILS_IMPORTER_PROVIDERS_ALLOWED_TO_DELETE_RECORDS"
-    ] and mode == 'delete':
+    ] and mode == 'DELETE':
```
ntarocco, line 61:
```suggestion
    ] and mode == ImporterMode.DELETE.value:
```
?
To be applied everywhere?

cds_ils/importer/api.py
```diff
@@ -43,61 +42,68 @@ def process_dump(data, provider, mode, source_type):
...
+                        log.id, record_recid,
+                        str(e.original_exception.message)
+                    )
                     continue
-
-                ImporterTaskEntry.create_success(entry_data, report)
     except Exception as e:
-        records_logger.error(
-            "@FILE TASK: {0} ERROR: {1}".format(log_id, str(e))
+        click.echo(
```
ntarocco, line 104: As discussed in the past, personally I don't love to use the `click.echo` where you are not sure that it will be only called by a CLI command. An API file might be called by other tasks I guess. Why remove the logger and use the click.echo?

cds_ils/importer/cli.py
```diff
@@ -42,97 +37,22 @@ def importer():
     "--mode",
     "-m",
     required=True,
-    type=click.Choice(["create", "delete"]),
+    type=click.Choice(["CREATE", "DELETE", "PREVIEW"]),
```
ntarocco, line 40: Use the enum here too?

cds_ils/importer/documents/api.py
```diff
@@ -85,14 +85,14 @@ def search_document_by_title_authors(title, authors, subtitle=None):
     document_search = current_app_ils.document_search_cls()
     if subtitle:
         search = (
-            document_search.query("match", title=title)
+            document_search.query("query_string", query=f"title:\"{title}\"")
```
ntarocco, line 88: Interesting change! Can you maybe quickly explain why from `match` to `query_string`?

ntarocco (approved):
I have a bit of hard time to review more than this without checking it out and test it, I am sorry if comment are not very useful...
I notice that there is no much change in the tests.
Please feel free to add more tests if you feel it is needed.

cds_ils/importer/tasks.py
```diff
@@ -51,7 +43,6 @@ def create_import_task(
 @shared_task
 def import_from_xml_task(log_id, source_path, source_type, provider, mode):
     """Load a single xml file task."""
-    # reset vocabularies validator cache
-    vocabulary_validator.reset()
-
-    import_from_xml(log_id, source_path, source_type, provider, mode)
+    log = ImporterTaskLog.query.get(log_id)
+    print(log)
```
ntarocco, line 47:
```suggestion
```

ui/src/importer/Importer.js
```diff
@@ -121,9 +132,10 @@ export class Importer extends Component {
           </Button>
           <Button
             primary
-            onClick={() => {
+            action="DELETE"
+            onClick={(e, props) => {
               this.setState({ openModal: false });
-              this.handleSubmit();
+              this.handleSubmit(e, props);
```
ntarocco, line 138: Why not passing only the `action` here? No need for `e` and all props if I am not mistaken.

ui/src/importer/importerTaskDetails/EitemImportDetails.js
```diff
@@ -0,0 +1,75 @@
...
+          <Button
+            color="red"
+            icon="exclamation"
+            floated="right"
+            size="mini"
+            basic
+          />
+        }
+      >
+        <Modal.Header>EItem - attention required</Modal.Header>
```
ntarocco, line 31: Why the text `attention required`?

ui/src/importer/importerTaskDetails/EitemImportDetails.js
```diff
@@ -0,0 +1,75 @@
...
+                    <List.Item key={pid}>
+                      <Link to={BackOfficeRoutes.eitemDetailsFor(pid)}>
+                        {pid}
+                      </Link>
+                    </List.Item>
+                  ))}
+                </List>
+              </>
+            )}
+            {/*<ReactJson src={jsonData} name={null} />*/}
```
ntarocco, line 62:
```suggestion
```

ui/src/importer/importerTaskDetails/EitemImportDetails.js
```diff
@@ -0,0 +1,75 @@
...
+export class EitemImportDetails extends Component {
+  constructor(props) {
+    super(props);
+    this.state = { open: false };
+  }
+  render() {
+    const { eitemReport } = this.props;
+    const { open } = this.state;
+    return (
+      <Modal
```
ntarocco, line 17: Not a fan of repeating modals inside for loops, these are extra components that are rendered, attached to the DOM, and it could be avoided.
Modals should be one unique only at the page level, that we fill, open and close as needed.

kprzerwa: I tried to find where is the for loop generating those modals, do you mean that we have multiple modals displaying JSON code, rendered in the table ?

ui/src/importer/importerTaskDetails/EitemImportDetails.js
kprzerwa: because it displays deleted or to be deleted eitems. I thought it might catch some more attention for this case.

ui/src/importer/Importer.js
kprzerwa: I did it because the code is cleaner in any other cases of assigning this method to onClick event:
```
              <Button
                secondary
                content="Import"
                onClick={this.handleSubmit}
                action="IMPORT"
              />
```
where signature `handleSubmit = (e, { action })` is quite neat. In this particular case however there is one more step needed to set the modal state before calling handleSubmit. So my options were to pass all the props of the button or to map it before method call. The former seemed cleaner to me in terms of props manipulation. Would you prefer I change it?

cds_ils/importer/tasks.py
kprzerwa: thanks, I noticed it as well, it is addressed in the next PR

tests/importer/test_importer.py
```diff
@@ -18,10 +18,11 @@ def test_modify_documents(importer_test_data):
...
     report = importer.import_record()
-    assert report["updated_document"]
+    assert report["document_json"]
+    assert report["action"] == "update"
 
     updated_document = document_cls.get_record_by_pid(
-        report["updated_document"]["pid"])
+        report["document_json"]["pid"])
     # wait for indexing
     time.sleep(1)
```
Pineirin, line 27: Why is it 1 second wait for indexing? Could it take more?

Pineirin (approved):

tests/importer/test_importer.py
kprzerwa: it is usually 1-2s, it never takes more than that, it was verified on different machines (otherwise the test wouldn't pass)

ui/src/importer/Importer.js
ntarocco: Ok I understand. The risk in general is that in the future someone will abuse of that, having all the props already there, and create some dirty code.
I would suggest to always first create the inner methods/funcs that will have, as params, only the minimum needed, nothing more (and ideally pure functions).

tests/importer/test_importer.py
ntarocco: What about using ES `flush` to wait instead?

ui/src/importer/Importer.js
kprzerwa: 👍  I will reimplement it

PR #549, opened 2021-07-01
global: add matomo analytics

kprzerwa:


ui/public/index.html
```diff
@@ -57,6 +57,23 @@
     </p>
   </noscript>
   <!-- End Piwik Tracking Code -->
+  <!-- Begin Matomo tracking Code -->
+  <script type="text/javascript">
```
ntarocco, line 61: No extra JS library to include? I guesss `_paq` has to be somewhere defined...
I would also suggest, if it works (to be tested), to move all the JS, including the `piwik` bundle `script` tag at the very bottom...

Pineirin (approved):

ui/public/index.html
kprzerwa: I checked twice and it does need any JS dependency

Issue #557, opened 2021-07-12
Importer Previewer - feedback report


agentilb:
### General comments

~~The first record is never displayed on the import page (but imported anyway). [BUG]~~
- [x] Blue preview button is not well displayed (Windows/Firefox, Mac/Firefox)
<img width="145" alt="Picture 1" src="https://user-images.githubusercontent.com/12181904/125264286-03ea6780-e304-11eb-99cc-cc345c526324.png">

- [x] #565
- [x] If the XML file is broken, the system should alert immediately and stop the process
- [x] A file with more than 1000 records breaks
<img width="279" alt="Picture 1" src="https://user-images.githubusercontent.com/12181904/125265608-4a8c9180-e305-11eb-8c07-abc7b7163b62.png">

### Preview
- [x] We should display more results on a page (=100)
- [ ] #571
- [x] The preview shows several times the same document [BUG]
- [ ] #563
- [ ] #564

### Import mode
~~Creation/replacement of e-item: the criteria used is the url.description, the internal note should be used instead.~~
- [ ] #572
- [ ] #566
- Could we maybe have a button to choose ‘yes, it is the same, update it’ or ‘no, it is not the same, create a new record’? 
- Otherwise, how could we force the import if not the same but the script keep indicating ‘partial matches’?

### Delete mode
- [ ] #567

### Details per Source

**EBL**

- [x] Adapt the rules, when indicator not known, ignore the indicator, and treat the field as indicated in the mapping
- [x] Remove EBC in the identifier (not added in the existing records)
- [x] Add in the internal note: EBLYYYYMM  (year and month of importation)
- [x] URL description: ebook
- [x] In case an EBL e-item is matched to a SPR e-item, it should be ignored. For the time being, it is added. 

**SPR**
- [x] Adapt the rules, when indicator not known, ignore the indicator, and treat the field as indicated in the mapping
- [x] Add in the internal note: SPRYYYYMM  (year and month of importation)
- [x] URL description: ebook

**SAF**
- [x] Adapt the rules, when indicator not known, ignore the indicator, and treat the field as indicated in the mapping
- [x] Add in the internal note: SAFYYYYMM  (year and month of importation)
- [x] URL description: ebook

kprzerwa:
```The first record is never displayed on the import page (but imported anyway). [BUG] ``` cannot reproduce, I checked and the first record displays fine.

agentilb:
Example:
[...]

PR #577, opened 2021-07-27
importer: cancel current import

### LINKED ISSUE #567: importer: deletion mode fixes
- If there is no item, no e-item, no series and no other dependancies, the record should also be deleted together with the e-item
- If there is no item, no e-item but other dependancies, the record should be higlighted so we can manually check the record

kprzerwa:
for reviewers, only 1 commit:
https://github.com/CERNDocumentServer/cds-ils/pull/577/commits/ea7c3c0b624a64429b5210c96ddca43160b3eaeb
* addresses #557 
* closes #567 

Screenshot 2021-07-27 at 14 57 18
Screenshot 2021-07-27 at 14 57 03
Screenshot 2021-07-27 at 14 57 23

### DIFF
--- a/cds_ils/importer/importer.py
+++ b/cds_ils/importer/importer.py
@@ -9,8 +9,12 @@
 import time
 
 from flask import current_app
+from invenio_app_ils.errors import RecordHasReferencesError
 from invenio_app_ils.proxies import current_app_ils
+from invenio_app_ils.records_relations.api import RecordRelationsParentChild
+from invenio_app_ils.relations.api import Relation
 from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
 
 from cds_ils.importer.documents.importer import DocumentImporter
 from cds_ils.importer.eitems.importer import EItemImporter
@@ -30,6 +34,7 @@ class Importer(object):
         "agency_code",
         "_serial",
         "provider_recid",
+        "_migration",
     )
 
     def __init__(self, json_data, metadata_provider):
@@ -39,6 +44,8 @@ class Importer(object):
         priority = current_app.config["CDS_ILS_IMPORTER_PROVIDERS"][
             metadata_provider
         ]["priority"]
+
+        eitem_json_data = self._extract_eitems_json()
         self.document_importer = DocumentImporter(
             json_data,
             self.HELPER_METADATA_FIELDS,
@@ -47,6 +54,7 @@ class Importer(object):
         )
         self.eitem_importer = EItemImporter(
             json_data,
+            eitem_json_data,
             metadata_provider,
             priority,
             self.IS_PROVIDER_PRIORITY_SENSITIVE,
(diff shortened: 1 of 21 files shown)


cds_ils/importer/views.py
```diff
@@ -76,15 +82,28 @@ def __init__(self, serializers=None, *args, **kwargs):
...
-        if log:
             return self.make_response(log, record_offset=offset)
-        else:
-            abort(404, "Task not found")
+        except ObjectDeletedError:
+            abort(404, "The task log was deleted.")
+        except NoResultFound:
+            abort(404, "Task does not exist.")
+
+    def post(self, log_id):
```
ntarocco, line 93: I think that you might have forgotten the permission check?
`@need_permissions("document-importer")`

cds_ils/importer/views.py
```diff
@@ -76,15 +82,28 @@ def __init__(self, serializers=None, *args, **kwargs):
...
+        """Cancels the ongoing celery task."""
+        try:
+            log = db.session.query(ImporterImportLog).get(log_id)
+            celery_task = AsyncResult(log.celery_task_id)
+            log.status = ImporterTaskStatus.CANCELLED
+            db.session.commit()
+            if not celery_task.ready():
+                celery_task.revoke()
```
ntarocco, line 101: I guess it is a nice to have? Isn't it better to let it exist on itself so that it will call `log.finalize`? It might revoked it before calling that...

ui/src/importer/importerTaskDetails/cancelImportTask.js
```diff
@@ -0,0 +1,89 @@
...
+      cancelLoading: false,
+      taskCancelled: undefined,
+    };
+  }
+
+  handleCancelImportAction = async () => {
+    const { logId } = this.props;
+    try {
+      this.setState({ cancelLoading: true });
+      const response = await importerApi.cancel(logId);
```
ntarocco, line 21: You should use `withCancel` pattern here to avoid memory leaks, given that you are not using Redux.

ui/src/importer/importerTaskDetails/cancelImportTask.js
```diff
@@ -0,0 +1,89 @@
...
+        <Button
+          color="red"
+          content="Cancel import"
+          onClick={this.handleOpenModal}
+        />{' '}
+        <Confirm
+          open={modalOpen}
+          onCancel={this.handleCloseModal}
+          onConfirm={this.handleCancelImportAction}
+          content="You are about to cancel current import. Are you sure you want to proceed?"
```
ntarocco, line 79: I would add something saying that the cancel operation will happen after the current import completed.

cds_ils/importer/importer.py
```diff
@@ -166,6 +181,38 @@ def delete_record(self):
         if self.document:
             self.output_pid = self.document["pid"]
             self.delete_records(self.document)
+            time.sleep(2)
```
ntarocco, line 184: use ES `flush` instead?

ntarocco (approved):

PR #583, opened 2021-08-20
ldap: fix remote account creation

kprzerwa:


cds_ils/ldap/user_importer.py
```diff
@@ -66,7 +66,7 @@ def create_invenio_remote_account(self, user_id, ldap_user):
         employee_id = ldap_user["remote_account_person_id"]
         department = ldap_user["remote_account_department"]
         mailbox = ldap_user["remote_account_mailbox"]
-        return RemoteAccount(
+        return RemoteAccount.create(
```
ntarocco, line 69: I am not sure that this will solve it, because the only thing it does is to add it to the session, which is already done below...

kprzerwa: it makes sure that the extra data is an empty dict, not null (null will cause an exception)

ntarocco (approved):

Issue #563, opened 2021-07-21
Import summary


kprzerwa:
Create import summary at the top of the results table:
XX records will be created, XX records will be updated, XX records with errors, etc.

PR #594, opened 2021-09-14
ImportedDocuments: display statistics of the import process.

Pineirin:
Display statistics during and after an import process of the imported records.
The displayed statistics are: records imported, records to be imported, records created, records updated, records with errors, records with eItem, records with serials.
closes https://github.com/CERNDocumentServer/cds-ils/issues/563

Newer new layout.
Screenshot from 2021-09-14 16-51-04
Also with blue variant, which has the problem of being possibly mistaken by a button.
Screenshot from 2021-09-14 16-51-21

New layout.
Screenshot from 2021-09-14 16-34-58

Old layout.
Screenshot from 2021-09-14 13-50-52

### DIFF
--- a/ui/src/importer/importerTaskDetails/ImportedDocuments.js
+++ b/ui/src/importer/importerTaskDetails/ImportedDocuments.js
@@ -3,19 +3,21 @@ import PropTypes from 'prop-types';
 import {
   Button,
   Divider,
-  Header,
   Icon,
   Message,
   Pagination,
   Table,
+  Segment,
+  Statistic,
+  Label,
 } from 'semantic-ui-react';
 import _isEmpty from 'lodash/isEmpty';
+import _isNull from 'lodash/isNull';
 import _get from 'lodash/get';
 import { CancelImportTask } from './cancelImportTask';
 import { EitemImportDetailsModal } from '../EitemImportDetailsModal';
 import { SeriesImportDetailsModal } from '../SeriesImportDetailsModal';
 import { JsonViewModal } from '../JsonViewModal';
-import { modeFormatter } from '../ImporterList';
 import { ImportedDocumentReport } from './ImportedDocumentReport';
 import { CdsBackOfficeRoutes } from '../../overridden/routes/BackofficeUrls';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


Pineirin:
Feedback in the demo.

    Change the import mode label as part of the statistics.
    Delete the title.

Pineirin:
I would still like to receive some further feedback on the layout. The IMPORT label seems huge, but it's not possible to reduce it using semantic ui while using the class that allow the the statistics to distribute themselves equally among the given space. 
A quite dirty solution could be to use as the value of the statistics field a label or an image, but I would like to hear your opinion on the topic before trying something like that.

ntarocco (commented):
It looks nice! Good job!!!
P.S. have a look to the linter warnings in the PR :) Most of them can be fixed.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -65,11 +72,47 @@ export class ImportedDocuments extends React.Component {
...
     } else {
       this.intervalId && clearInterval(this.intervalId);
     }
   };
 
+  calculateStatistics = data => {
+    const importStatistics = [];
+    importStatistics.push({
+      text: 'Mode',
+      value: <div className="ui label">{data.mode}</div>,
```
kprzerwa, line 85: this is not the correct way of using semantic ui react label component

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -65,11 +72,47 @@ export class ImportedDocuments extends React.Component {
...
+    });
+    importStatistics.push({
+      text: 'Records',
+      value: data.loaded_entries + '/' + data.entries_count,
+    });
+    importStatistics.push({
+      text: 'Records created',
+      value: data.records.filter(record => record.action === 'create').length,
```
kprzerwa, line 93: is there any more efficient way than filtering all the records each time the data is refreshed ? Did you try on a file with 500-1000+ records, if the UI will be performing well and fast ?

Pineirin: I will address this issue in a separated PR, as it involves important code changes and is quite bound to the filter process.

Pineirin: The work is being done in this PR.
https://github.com/CERNDocumentServer/cds-ils/pull/603

PR #599, opened 2021-09-21
Importer: add  filter to the importer results.

Pineirin:
Add the posibility for the user to filter the importer results table based on the different criteria displayed on the statistics fields.
closes https://github.com/CERNDocumentServer/cds-ils/issues/564

### DIFF
--- a/ui/src/importer/importerTaskDetails/ImportedTable.js
+++ b/ui/src/importer/importerTaskDetails/ImportedTable.js
@@ -1,0 +1,120 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Pagination, Table } from 'semantic-ui-react';
+import _isEmpty from 'lodash/isEmpty';
+import { EitemImportDetailsModal } from '../EitemImportDetailsModal';
+import { SeriesImportDetailsModal } from '../SeriesImportDetailsModal';
+import { JsonViewModal } from '../JsonViewModal';
+import { ImportedDocumentReport } from './ImportedDocumentReport';
+
+export class ImportedTable extends React.Component {
+  constructor(props) {
+    super(props);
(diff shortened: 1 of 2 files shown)


ui/src/importer/importerTaskDetails/ImportedTable.js
```diff
@@ -0,0 +1,122 @@
...
+import { EitemImportDetailsModal } from '../EitemImportDetailsModal';
+import { SeriesImportDetailsModal } from '../SeriesImportDetailsModal';
+import { JsonViewModal } from '../JsonViewModal';
+import { ImportedDocumentReport } from './ImportedDocumentReport';
+
+export class ImportedTable extends React.Component {
+  constructor(props) {
+    super(props);
+    this.pageSize = 100;
+    const { records } = this.props;
```
Pineirin, line 14: I really don't like storing the previous records to see if I should reset to page 1, but I run out of ideas so maybe I can rethink it with a bit of feedback. I don't know either if there is a way to destructure the records inside an object in which I have other fixed values, but I wasn't able.

ui/src/importer/importerTaskDetails/ImportedTable.js
```diff
@@ -0,0 +1,122 @@
...
+    this.state = {
+      activePage: 1,
+      records: records,
+    };
+  }
+
+  static getDerivedStateFromProps(props, state) {
```
Pineirin, line 21: As pointed previously, I didn't manage to find a better way to update the activePage (and with that the state) every time the records changed. Getting the activePage out of the state isn't a good solution also, as it allows to update the page with each page change. Regardless if I receive feedback or not I will update this code, as the check is open for bugs if both records are different but have the same length.

PR #603, opened 2021-09-22
Import Documents: statistics and filtering

Pineirin:
Display statistics during and after an import process of the imported records.
The displayed statistics are: records imported, records to be imported, records created, records updated, records with errors, records with eItem, records with serials.
Add the possibility for the user to filter the importer results table based on the different criteria displayed on the statistics fields.
closes https://github.com/CERNDocumentServer/cds-ils/issues/563
part of https://github.com/CERNDocumentServer/cds-ils/pull/599
Default page.
Screenshot from 2021-09-23 10-17-04
Filter by records with serials.
Screenshot from 2021-09-23 10-17-12
Filter by records with errors.
Screenshot from 2021-09-23 10-17-25

Previous discussion in this PRs:
https://github.com/CERNDocumentServer/cds-ils/pull/594.
https://github.com/CERNDocumentServer/cds-ils/pull/599

### DIFF
--- a/ui/src/importer/importerTaskDetails/ImportedDocuments.js
+++ b/ui/src/importer/importerTaskDetails/ImportedDocuments.js
@@ -3,19 +3,20 @@ import PropTypes from 'prop-types';
 import {
   Button,
   Divider,
-  Header,
   Icon,
   Message,
   Pagination,
   Table,
+  Segment,
+  Statistic,
 } from 'semantic-ui-react';
 import _isEmpty from 'lodash/isEmpty';
+import _isNull from 'lodash/isNull';
 import _get from 'lodash/get';
 import { CancelImportTask } from './cancelImportTask';
 import { EitemImportDetailsModal } from '../EitemImportDetailsModal';
 import { SeriesImportDetailsModal } from '../SeriesImportDetailsModal';
 import { JsonViewModal } from '../JsonViewModal';
-import { modeFormatter } from '../ImporterList';
 import { ImportedDocumentReport } from './ImportedDocumentReport';
 import { CdsBackOfficeRoutes } from '../../overridden/routes/BackofficeUrls';
 import { Link } from 'react-router-dom';
@@ -31,6 +32,7 @@ export class ImportedDocuments extends React.Component {
       data: null,
       isLoading: true,
       activePage: 1,
+      statistics: [],
     };
   }
 
@@ -53,6 +55,11 @@ export class ImportedDocuments extends React.Component {
     if (!importCompleted) {
       const nextEntry = _get(data, 'loaded_entries', 0);
       const response = await importerApi.check(taskId, nextEntry);
+      const responseData = response.data;
+
+      if (responseData) {
+        responseData.records = _get(responseData, 'records', []);
+      }
       if (response.data.status !== 'RUNNING') {
         this.setState({
           importCompleted: true,
@@ -65,9 +72,45 @@ export class ImportedDocuments extends React.Component {
           isLoading: true,
         });
       }
+      this.calculateStatistics(response.data);
     } else {
       this.intervalId && clearInterval(this.intervalId);
     }
+  };
+
+  calculateStatistics = data => {
+    const importStatistics = [];
+    importStatistics.push({
+      text: 'Mode',
+      value: <div className="ui label">{data.mode}</div>,
+    });
+    importStatistics.push({
+      text: 'Records',
+      value: data.loaded_entries + '/' + data.entries_count,
+    });
+    importStatistics.push({
+      text: 'Records created',
+      value: data.records.filter(record => record.action === 'create').length,
+    });
+    importStatistics.push({
+      text: 'Records updated',
+      value: data.records.filter(record => record.action === 'update').length,
+    });
+    importStatistics.push({
+      text: 'Records with errors',
+      value: data.records.filter(record => _isNull(record.action)).length,
+    });
+    importStatistics.push({
+      text: 'Records with eItem',
+      value: data.records.filter(record => !_isNull(record.eitem)).length,
+    });
+    importStatistics.push({
+      text: 'Records with Serials',
+      value: data.records.filter(record => !_isEmpty(record.series)).length,
+    });
+    this.setState({
+      statistics: importStatistics,
+    });
   };
 
   renderErrorMessage = () => {
@@ -132,6 +175,23 @@ export class ImportedDocuments extends React.Component {
 
   handlePaginationChange = (e, { activePage }) => this.setState({ activePage });
 
+  renderStatistics = () => {
+    const { statistics } = this.state;
+
+    return (
+      <Statistic.Group widths="seven">
+        {statistics.map(function(statistic) {
+          return (
+            <Statistic key={statistic.text}>
+              <Statistic.Value>{statistic.value}</Statistic.Value>
+              <Statistic.Label>{statistic.text}</Statistic.Label>
+            </Statistic>
+          );
+        })}
+      </Statistic.Group>
+    );
+  };
+
   renderResultsContent = () => {
     const { data, activePage } = this.state;
     return (
@@ -187,7 +247,7 @@ export class ImportedDocuments extends React.Component {
                 return (
                   !_isEmpty(elem) && (
                     <ImportedDocumentReport
-                      key={index}
+                      key={elem.output_pid}
                       documentReport={elem}
                       listIndex={index + (activePage - 1) * this.pageSize + 1}
                     />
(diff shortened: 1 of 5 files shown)


ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
...
+      this.modeLabel();
+      this.statusLabel();
+      this.calculateStatistics();
     } else {
       this.intervalId && clearInterval(this.intervalId);
     }
   };
 
+  modeLabel = () => {
+    let { data, mode } = this.state;
```
ntarocco, line 121:
```suggestion
    const { data, mode } = this.state;
```

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
...
+        mode = (
+          <Label color="teal" basic>
+            Preview (delete)
+          </Label>
+        );
+        break;
+      default:
+        mode = (
+          <Label color="black" basic>
+            Invalid mode
```
ntarocco, line 154: We do need a default, but it is probably not very nice to show "Invalid mode". Often, in the code, we raise an exception is such cases. Try to check other `default:` in other `switch` cases.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
...
+        );
+        break;
+    }
+    this.setState({
+      mode: mode,
+    });
+  };
+
+  statusLabel = () => {
+    let { data, status, isLoading } = this.state;
```
ntarocco, line 165:
```suggestion
    const { data, isLoading } = this.state;
    let status;
```

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
...
+      : data.status === 'CANCELLED'
+      ? (status = <Label color="yellow">Import cancelled</Label>)
+      : (status = <Label color="red">Import failed</Label>);
+    this.setState({
+      status: status,
+    });
+  };
+
+  calculateStatistics = () => {
+    let { statistics, data, selectedResult } = this.state;
```
ntarocco, line 185:
```suggestion
    const { statistics, data, selectedResult } = this.state;
```

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -25,13 +23,54 @@ import { invenioConfig } from '@inveniosoftware/react-invenio-app-ils';
...
+        records: {
+          text: 'Records',
+          value: '',
+          filterFunction: a => a,
+          filteredRecords: [],
+        },
+        records_created: {
+          text: 'Records created',
+          value: '',
+          filterFunction: a => a.action === 'create',
```
ntarocco, line 41: Let's avoid naming variables `a`, `b`, `c` if possible ;)

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
...
+    }
+    this.setState({
+      mode: mode,
+    });
+  };
+
+  statusLabel = () => {
+    let { data, status, isLoading } = this.state;
+    !data
+      ? (status = <Label color="grey">Fetching status</Label>)
```
ntarocco, line 167: I am still trying to understand if we can improve code readability here. My feeling is that, since we check the `data` and the `data.status`, I wouldn't mix the things.
```
if (!data) { // or !!data?
  status = <Label color="grey">Fetching status</Label>
} else {
  switch...
    RUNNING
    SUCCEEDED
    FAILED
    CANCELLED
}
```
Since the content of these cmps is static, it doesn't depend on the status, I would avoid to re-create each time.
You could create some consts on the top of the cmp:
```
STATUS_CMPS = {
  'RUNNING' = <Label...,
  '...'
}
```
and then use it in the render method:
```
cmp = STATUS_CMPS[data.status];
```

Same for the mode above. You would avoid also to have a *fat* state by putting much less cmps inside.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
...
+    }
+    this.setState({
+      statistics: statistics,
+    });
+    this.currentFilterResults(selectedResult);
+  };
+
+  currentFilterResults(key) {
+    const { statistics } = this.state;
+    this.setState({
```
ntarocco, line 215: I am wondering if it would be better to calculate this in the `render` method instead of having a part of the state with all the records, and another part with the filtered. If we have e.g. 1K records, you might have 2 times these 1K records in the state.
We can probably chat about this together.

ntarocco (commented):
Nice work! Looks great! :)
I have added some comments, but it is probably worth it to chat about the solutions before making changes :)

ui/src/importer/importerTaskDetails/ImportedDocuments.js
kprzerwa: I agree, explicit naming would be more readable

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -53,23 +92,132 @@ export class ImportedDocuments extends React.Component {
     if (!importCompleted) {
       const nextEntry = _get(data, 'loaded_entries', 0);
       const response = await importerApi.check(taskId, nextEntry);
+      const responseData = response.data;
+
+      if (responseData) {
```
kprzerwa, line 97: I removed this part before, mainly because it was cause huge problems with pagination, could you test if it works properly ?
Also maybe it is an optimisation for the next PR, but by setting
```
responseData.records = _get(responseData, 'records', []
```
you are just overwriting all the data each time it is refreshed, instead of adding small chunks. The initial problem is that the backend endpoint is returning all the records all the time, instead of respecting the `offset` parameter. If you would have to handle only few records at each refresh, it would be even more justified, to keep each filter under separate state as you do above
ping @ntarocco

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -97,163 +245,92 @@ export class ImportedDocuments extends React.Component {
...
     );
   };
 
   renderCancelButton = () => {
     const { data } = this.state;
     const { taskId } = this.props;
+    const isRunning = !_isEmpty(data) && data.status === 'RUNNING';
 
-    return (
+    return isRunning ? (
```
kprzerwa, line 281: you could use `&& ` operator instead of `? : null`

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -97,163 +245,92 @@ export class ImportedDocuments extends React.Component {
...
+                  <Grid className="middle aligned">
+                    <Grid.Column width={3}>
+                      <Segment>
+                        Mode: {mode}
+                        <Divider />
+                        Status: {status}
+                      </Segment>
+                    </Grid.Column>
+                    <Grid.Column width={13}>
+                      {this.renderStatistics()}
```
kprzerwa, line 316: does it make sense to have a separate component here instead of method? I would like to avoid the JSX code spread around in the methods different than render if possible

ui/src/importer/importerTaskDetails/ImportedTable.js
```diff
@@ -0,0 +1,120 @@
...
+              .slice(
+                activePage * this.pageSize - this.pageSize,
+                activePage * this.pageSize
+              )
+              .map((elem, index) => {
+                return (
+                  !_isEmpty(elem) && (
+                    <ImportedDocumentReport
+                      // Keep the index as key due to lack of other unique id.
+                      key={index}
```
kprzerwa, line 87: you could add a prettier exception to suppress this warning

ui/src/importer/importerTaskDetails/cancelImportTask.js
```diff
@@ -80,15 +80,12 @@ export class CancelImportTask extends Component {
...
           onCancel={this.handleCloseModal}
           onConfirm={this.handleCancelImportAction}
-          content={
-            <>
-              You are about to cancel current import - the cancel will be
-              performed after the current record is processed.
-              <strong>Are you sure you want to proceed?</strong>
-            </>
-          }
+          header="Import cancel."
```
kprzerwa, line 83: 👍

ui/src/importer/importerTaskDetails/ImportedDocuments.js
kprzerwa: I think if we don't do the filtering on the backend, maybe its better to keep the 2k in the state instead of filter on each render (which in the end will be either way 2k kept in the memory, just different scopes), also for the reasons written in https://github.com/CERNDocumentServer/cds-ils/pull/603#discussion_r719395116

ui/src/importer/importerTaskDetails/ImportedDocuments.js
NRodriguezcuellar: another alternative to if / else is using guard clauses so just return it early if a a certain condition is met so that you do not need to wrap it in an else statement 

```
if(!data) {
return <Label .....>
}

switch () {
......
}

```

PR #624, opened 2021-10-12
EItem: Add aditional sources.

Pineirin:
* Add the sources proposed by the librarians to the ones already available when creating an EItem.
* closes: CERNDocumentServer#607
Screenshot from 2021-10-12 10-26-23

### DIFF
--- a/cds_ils/vocabularies/data/eitem_sources.json
+++ b/cds_ils/vocabularies/data/eitem_sources.json
@@ -9,9 +9,74 @@
     "type": "eitem_sources",
     "text": "O'Reilly"
   },
-    {
+  {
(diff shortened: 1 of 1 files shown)


cds_ils/migrator/eitems/api.py
```diff
@@ -94,7 +94,20 @@ def add_eitem_extra_metadata(eitem, document):
     PROVIDERS_MAPPING = {
         "safari": "SAF",
         "springer": "SPR",
-        "ebl": "EBL"
+        "ebl": "EBL",
```
kprzerwa, line 97: please remove this part, as it was meant for migration, and these providers do not appear in the migrated data

Issue #608, opened 2021-09-27
import delete: display issues when deleting records with dependencies


kprzerwa:
from the user feedback:

> https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/118
No 2 and 32: no error shown in the preview, but error in the import. Document not updated (the EBC identifier should be removed). E-item well deleted but not showed in the import.

> Difference Preview / Import : https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/153 vs. https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/155 -> The documents are marked as "update" in the preview and "delete" in the import: this is fine because the import is correct, but still confusing not to have the same message in both modes.

> Difference Preview / Import : https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/153 vs. https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/155 -> Line 4 has no error in the preview and an error in the import

> https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/155 line 192 marked as 'update' but no link to the document. If document not found, it should be marked as non matched.

task: 
- [x] show the deleted eitem on errored deletes(when document cannot be deleted)
- [x]  show in the preview delete label on the document if document can be safely deleted
- [x] show error if document has dependencies (on delete preview)
- [x] fix update labels in not matched documents
- [x] display pid of deleted eitem both in preview and real delete.
- [x] display pid of created series in real import mode.

PR #646, opened 2021-10-29
Importer: fix inconsistency in preview statuses 

Pineirin:
Fix "has only serial relations" check. In the past there was a check for making sure a record had only serial relations (if any), but, instead of checking that all the relations that it had were serial, it was checking that it had relations AND they were serial. This was fixed to aim for it's true purpose.

Raise errors in preview delete. In the past, in preview delete mode, we would mark a record as "delete" if it didn't had any non serial relations; meanwhile in real delete, we would try to delete it anyways and raise and exception if there was any problem due to this relations, now the behaviour in the preview delete has been changed so it can raise the same exceptions. 

Fix inconsistency in preview statuses. In the past, in delete mode, the importer would try to delete the record even if it had any non serial relations (in case of success the record would be marked as "delete"); meanwhile, in preview delete mode, the importer would mark the record as delete if it didn't have any non-serial relations. Now, all records are marked to "delete" in preview mode (if they exist), unless there is an exception, in which case they will be marked as "error", just like in delete mode.

Add none label for unmatched deletions. The importer can try to delete records that don't exist in the database, previously this unmatched records were marked as "update", now they are marked as "none".

Hide eItem rawJson when there isn't eItem. In the past, if a record had an empty eItem, a button to display it's raw json would appear in the UI. Now it doesn't.

Properly calculate statistics of eItem and series. In the past non null but empty eItems where taken into account when calculating the statistics. Now they aren't.

Display pid in series creation. In the past, when a new series was created for a record, the PID of the new series wasn't stored in the database, so it wasn't displayed in the UI. Now it's properly stored.

closes https://github.com/CERNDocumentServer/cds-ils/issues/608
Screenshot from 2021-11-03 10-49-18
Screenshot from 2021-11-03 10-49-11

### DIFF
--- a/cds_ils/importer/errors.py
+++ b/cds_ils/importer/errors.py
@@ -9,6 +9,7 @@
 """Exceptions."""
 
 from dojson.errors import DoJSONException
+from invenio_app_ils.errors import RecordHasReferencesError
 
 
 class LossyConversion(DoJSONException):
@@ -57,6 +58,20 @@ class CDSImporterException(DoJSONException):
         self.description = self.message
 
         super(CDSImporterException, self).__init__(*args)
+
+
+class DocumentHasReferencesError(RecordHasReferencesError):
+    """DocumentHasReferencesError class."""
+
+    def __init__(self, document, ref_type, refs):
+        """Constructor."""
+        ref_ids = sorted([res["pid"] for res in refs.scan()])
+        super().__init__(
+            record_type="Document",
+            record_id=document["pid"],
+            ref_type=ref_type,
+            ref_ids=ref_ids,
+        )
 
 
 class RecordModelMissing(CDSImporterException):
--- a/cds_ils/importer/importer.py
+++ b/cds_ils/importer/importer.py
@@ -16,12 +16,15 @@ from invenio_app_ils.records_relations.api import RecordRelationsParentChild
 from invenio_app_ils.relations.api import Relation
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier
+from invenio_records.errors import MissingModelError
 from invenio_search import current_search
 
 from cds_ils.importer.documents.importer import DocumentImporter
 from cds_ils.importer.eitems.importer import EItemImporter
 from cds_ils.importer.errors import UnknownProvider
 from cds_ils.importer.series.importer import SeriesImporter
+
+from .errors import DocumentHasReferencesError
 
 
 class Importer(object):
@@ -188,20 +191,21 @@ class Importer(object):
         document_indexer = current_app_ils.document_indexer
         series_class = current_app_ils.series_record_cls
         self._validate_provider()
-        self.action = "update"
+        self.action = "none"
         # finds the exact match, update records
         self.document = self._match_document()
 
         if self.document:
             self.output_pid = self.document["pid"]
             self.delete_records(self.document)
+            db.session.commit()
             current_search.flush_and_refresh(index="*")
             document_has_only_serial_relations = \
                 len(self.document.relations.keys()) \
                 and 'serial' in self.document.relations.keys()
 
             if not self.document.has_references() \
-                    and document_has_only_serial_relations:
+                    or document_has_only_serial_relations:
 
                 # remove serial relations
                 rr = RecordRelationsParentChild()
@@ -243,13 +247,13 @@ class Importer(object):
             "series": self.series_list,
             "raw_json": self.json_data,
             "document_json": doc_json,
-            "document": self.document
+            "document": self.document,
         }
 
     def preview_delete(self):
         """Preview deleting a record."""
         self._validate_provider()
-        self.action = "update"
+        self.action = "none"
         # finds the exact match, update records
         self.document = self._match_document()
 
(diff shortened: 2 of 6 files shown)


ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -61,12 +61,14 @@ export class ImportedDocuments extends React.Component {
         records_with_item: {
           text: 'Records with eItem',
           value: 0,
-          filterFunction: record => !_isNull(record.eitem),
+          filterFunction: record =>
+            !_isNull(record.eitem) && !_isEmpty(record.eitem),
```
kprzerwa, line 65: isn't the isEmpty check enough? (to be tested)

cds_ils/importer/importer.py
```diff
@@ -243,26 +244,22 @@ def import_summary(self):
...
-        self.action = "update"
+        self.action = "none"
         # finds the exact match, update records
         self.document = self._match_document()
 
         if self.document:
             self.output_pid = self.document["pid"]
             self.eitem_summary = \
                 self.eitem_importer.preview_delete(self.document)
-            document_has_only_serial_relations = \
```
kprzerwa, line 260: this check can be removed only if we raise an exception for "multipart_monograph" relation reference, which I think it is not added below

cds_ils/importer/errors.py
```diff
@@ -59,6 +60,20 @@ def __init__(self, *args, **kwargs):
         super(CDSImporterException, self).__init__(*args)
 
 
+class CDSRecordHasReferencesError(RecordHasReferencesError):
```
kprzerwa, line 63:
```suggestion
class DocumentHasReferencesError(RecordHasReferencesError):
```

cds_ils/importer/importer.py
```diff
@@ -288,3 +288,62 @@ def preview_import(self):
...
+                    related_refs.add("{pid_value}:{pid_type}".format(**obj))
+        if related_refs:
+            raise RecordHasReferencesError(
+                record_type=document.__class__.__name__,
+                record_id=document["pid"],
+                ref_type="related",
+                ref_ids=sorted(ref for ref in related_refs),
+            )
+
+        if document.model is None:
```
kprzerwa, line 348: not sure why you are checking the model here ?

Pineirin: It's just another check that happens when you try to delete an item so, if it can happen in the real delete, it should also be possible in the preview.

kprzerwa: not sure if I understand what you mean, lets discuss in person

cds_ils/importer/importer.py
```diff
@@ -243,26 +247,22 @@ def import_summary(self):
...
             self.eitem_summary = \
                 self.eitem_importer.preview_delete(self.document)
-            document_has_only_serial_relations = \
-                len(self.document.relations.keys()) \
-                and 'serial' in self.document.relations.keys()
-            if not self.document.has_references() \
-                    and document_has_only_serial_relations:
-                self.action = "delete"
+            self.preview_delete_document(self.document)
+            self.action = "delete"
```
zzacharo, line 265: @kprzerwa is it fine to just assign the action here regardless? I saw there were some conditions before that were applied in order to do that.

cds_ils/importer/importer.py
zzacharo: I am not sure about this either :/

ui/src/importer/importerTaskDetails/ImportedDocumentReport.js
```diff
@@ -17,6 +17,8 @@ class ImportedDocumentReportComponent extends Component {
       color = 'green';
     } else if (action === 'update' || action === 'replace') {
       color = 'yellow';
+    } else if (action === 'none') {
```
zzacharo, line 20: action is none in case preview only? If so, why don't we set the action value to `preview`?

kprzerwa: because the preview is not an action in this context. None = nothing will be done with the record (f.e. when we are trying to delete a record, but the record does not exist in the db, the action will be none, either you run a preview or an actual import)

cds_ils/importer/importer.py
Pineirin: In the delete we try to delete even if the condition is not fulfilled (probably raising an error). As the results between the preview and the real delete must be the same, the behaviour here was changed to meet that objective.

ui/src/importer/importerTaskDetails/ImportedDocumentReport.js
Pineirin: Action is none if the document was not found in the dabatase, either in real or in preview delete.

cds_ils/importer/importer.py
Pineirin: My rational here is the following:
When we try to delete a document we call to the delete method of invenio_app_ils/documents/api/Document which at the end calls the delete function of it's parent invenio_app_ils/records_relations/api/IlsRecordWithRelations which also calls in the end the function method of it's parent invenio_records/api/Record. In this last delete function there is a condition stating that if the model of the record is None a MissingModelError exception should be raised.
I don't think it could happen in our case but, if we want the preview to make the same checks than the real delete and raise the same errors, I think this should be added.
Regardless, this is just what I think and I don't feel confident stating this should be kept.

kprzerwa: lets discuss this in a call as soon as possible

cds_ils/importer/importer.py
```diff
@@ -288,3 +288,349 @@ def preview_import(self):
             self.action = None
 
         return self.import_summary()
+# -*- coding: utf-8 -*-
```
kprzerwa, line 291: I think you have a big issue with rebase here

kprzerwa (approved):

zzacharo (approved):

PR #649, opened 2021-11-01
Importer Task Detail: code refactor and error on unknown taskId

NRodriguezcuellar:
The Importer task detail's page was not showing errors when the request failed. Now it does.

before:

Screenshot 2021-10-27 at 14 49 27

After

Screenshot 2021-10-28 at 14 48 26
:


ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -288,7 +215,8 @@ export class ImportedDocuments extends React.Component {
   };
 
   filteredRecords = () => {
```
zzacharo, line 217: Cosmetic: Can you please rename this to `filterRecords` instead as the name refer to the action that this functions does?

ui/src/importer/importerTaskDetails/ImporterReportHeader.js
```diff
@@ -0,0 +1,47 @@
...
+    const { status, isLoading, taskId, onSearch } = this.props;
+    return (
+      <Grid>
+        <Grid.Column floated="left" width={12}>
+          <Button
+            className="default-margin-top"
+            labelPosition="left"
+            icon="left arrow"
+            content="Import other files"
+            loading={status}
```
zzacharo, line 20: Why are you not using the `isLoading` prop here?

ui/src/importer/importerTaskDetails/ImporterReportHeader.js
```diff
@@ -0,0 +1,47 @@
...
+            disabled={status}
```
zzacharo, line 21: Maybe change this to `disabled` instead of status? It reflects better what the passed property is controlling.

zzacharo (commented):
Nice job! Some cosmetics from my iteration!

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -76,48 +68,19 @@ export class ImportedDocuments extends React.Component {
...
-    } else {
-      this.intervalId && clearInterval(this.intervalId);
+      this.calculateStatistics(data);
     }
-  };
+  }
 
   calculateMode = () => {
-    const { data } = this.state;
+    const { data } = this.props;
```
kprzerwa, line 83: I would move this to a functional component

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -160,7 +123,7 @@ export class ImportedDocuments extends React.Component {
   };
 
   statusLabel = () => {
-    const { data, isLoading } = this.state;
+    const { data, isLoading } = this.props;
```
kprzerwa, line 126: this function as well. it is a part of presentation

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -76,48 +68,19 @@ export class ImportedDocuments extends React.Component {
     this.lastIndex = 0;
```
kprzerwa, line 68:
```
this.mode = '';
this.status = '';
this.lastIndex = 0;
```
these 3 variables should not be stored as class attribute, I know it was there before but could you change it to be stored in the component state ?

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportHeader.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return (
+      <Grid>
+        <Grid.Column floated="left" width={12}>
+          <Button
+            className="default-margin-top"
+            labelPosition="left"
+            icon="left arrow"
+            content="Import other files"
+            loading={isDisabled}
```
zzacharo, line 21: Is this really necessary to be set? If so, `isLoading` shouldn't be the value to control it?

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -301,83 +153,93 @@ export class ImportedDocuments extends React.Component {
...
+          isLoading={isLoading}
+          onSearch={this.onSearchChange}
+          isDisabled={!data?.status}
+        />
+
+        {entriesReady && dataAvailable ? (
+          <Grid className="middle aligned">
+            <Grid.Column width={3}>
+              <Segment>
+                Mode:
```
zzacharo, line 202: Maybe this should be part of the `ImporterReportMode` component.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -301,83 +153,93 @@ export class ImportedDocuments extends React.Component {
...
+                <ImporterReportMode data={data} />
+                <Divider />
+                Status:
```
zzacharo, line 205: Same here

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportHeader.js
NRodriguezcuellar: isLoading is a bit different, i tried both and using isLoading is not the wanted behaviour

NRodriguezcuellar: the reason why i don't think i want to wrap my head around it too much

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportMode.js
```diff
@@ -0,0 +1,43 @@
...
+          Preview (delete)
+        </Label>
+      ),
+      ERROR: (
+        <Label color="red" basic>
+          Error
+        </Label>
+      ),
+    };
+
```
Pineirin, line 36: Maybe I'm completely wrong here, but I think it's bad idea to have non needed code in the render because it only can use one thread. Maybe it would be better to have the labels object in the constructor?

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportStatusLabel.js
```diff
@@ -0,0 +1,35 @@
...
+      SUCCEEDED: <Label color="green">Import successful</Label>,
+      CANCELLED: <Label color="yellow">Import cancelled</Label>,
+      IMPORTING: (
+        <Label color="blue">
+          <Icon loading name="circle notch" /> Importing literature
+        </Label>
+      ),
+      LOADING: <Label color="grey">Fetching status</Label>,
+      FAILED: <Label color="yellow">Import failed</Label>,
+    };
```
Pineirin, line 19: Same comment as above.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -70,125 +56,26 @@ export class ImportedDocuments extends React.Component {
...
-        </Label>
-      );
-    switch (data.status) {
-      case 'SUCCEEDED':
-        return <Label color="green">Import successfull</Label>;
-      case 'CANCELLED':
-        return <Label color="yellow">Import cancelled</Label>;
-      default:
-        return <Label color="red">Import failed</Label>;
+    if (data && data !== prevProps.data) {
```
Pineirin, line 66: Ping @ntarocco. Maybe I forgot, but I remember something about only comparing a current property to the previous properties as a last resort.

Pineirin (commented):
Please remove all the warnings (you can check them in the files changed page of the PR).

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportMode.js
kprzerwa: I am not sure what do you mean by `render because it only can use one thread`, could you explain ? Could be in person.
However I agree with the comment, there is no reason to reinitialise the label const each time we re-render the component, since the label mapping is "static"

kprzerwa (commented):
this component became much more readable! thanks for refactoring!
only minor comments below

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportMode.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Label } from 'semantic-ui-react';
+
+export class ImporterReportMode extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
kprzerwa, line 9: why do you keep it in the state? It will not change

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportStatusLabel.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Icon, Label } from 'semantic-ui-react';
+
+export class ImporterReportStatusLabel extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
kprzerwa, line 9: same here

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -301,83 +153,91 @@ export class ImportedDocuments extends React.Component {
...
+
+    const dataAvailable = !_isEmpty(data);
+    const filteredRecordsAvailable = !_isEmpty(filteredRecords);
+    const filteredRecordsNotNull = !_isNull(filteredRecords);
+    const errorWhileFetching = !_isEmpty(data) && data?.status === 'FAILED';
+    const entriesReady =
+      (data?.loaded_entries || data?.loaded_entries === 0) &&
+      data?.entries_count;
+
+    if (error) {
```
kprzerwa, line 173: the not found will be displayed on any error in this case, shouldn't be only for 404?

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -70,125 +56,26 @@ export class ImportedDocuments extends React.Component {
...
-      );
-    switch (data.status) {
-      case 'SUCCEEDED':
-        return <Label color="green">Import successfull</Label>;
-      case 'CANCELLED':
-        return <Label color="yellow">Import cancelled</Label>;
-      default:
-        return <Label color="red">Import failed</Label>;
+    if (data && data !== prevProps.data) {
+      this.filterRecords();
```
ntarocco, line 67: Yes, thanks @Pineirin, I would avoid as much as possible to use `componentDidUpdate` and then compare props. 
It can lead to nasty bugs.
Why is this needed? Do you need it at first run?

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -70,125 +56,26 @@ export class ImportedDocuments extends React.Component {
...
   calculateStatistics = data => {
-    const { statistics } = this.state;
+    const { statistics, lastIndex } = this.state;
     let stats = _cloneDeep(statistics);
     const records = data.records;
-    const newRecords = records.slice(this.lastIndex, records.length);
+    const newRecords = records.slice(lastIndex, records.length);
 
-    this.lastIndex += newRecords.length;
+    this.setState({ lastIndex: lastIndex + newRecords.length });
```
ntarocco, line 78: I would move this down to the final `setState` before returning, to have only one `setState`: normally it should not immediately re-render, but @Pineirin observed that it was re-rendering on each `setState`.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
NRodriguezcuellar: That is true but i thought about some errors that could happen and i felt like it could serve multiple purposes, it just simplifies the code and logic.

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -301,83 +153,91 @@ export class ImportedDocuments extends React.Component {
...
-      data,
       statistics,
       selectedResult,
       filteredRecords,
       activePage,
     } = this.state;
+
+    const dataAvailable = !_isEmpty(data);
+    const filteredRecordsAvailable = !_isEmpty(filteredRecords);
+    const filteredRecordsNotNull = !_isNull(filteredRecords);
```
ntarocco, line 167: This is curious. Why not simply have one variable `const filteredRecordsAvailable = !_isEmpty(filteredRecords);`
and then check if there are no results with `!filteredRecordsAvailable`?

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportStatusLabel.js
```diff
@@ -0,0 +1,42 @@
...
+      ),
+      LOADING: <Label color="grey">Fetching status</Label>,
+      FAILED: <Label color="yellow">Import failed</Label>,
+    };
+  }
+
+  relevantLabel = status => {
+    const statusExists = !!this.labels[status];
+
+    return statusExists ? this.labels[status] : this.labels['FAILED'];
```
zzacharo, line 25: My only concern is here: We should understand when the status is missing and if it's because of failure, the backend should generate a status. Otherwise, for any potential bug we will end up to the failed label. This is fine from the UI perspective but it just covers the underlying bug! @NRodriguezcuellar could you elaborate when did you notice that the status is empty?

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -301,83 +229,91 @@ export class ImportedDocuments extends React.Component {
...
+    const errorWhileFetching = !_isEmpty(data) && data?.status === 'FAILED';
```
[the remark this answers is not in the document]
ntarocco: Yes but what happens then when there is an error from the start? `data` is empty  and status is `FAILED`?

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportStatusLabel.js
NRodriguezcuellar: That is a fair point! I have not noticed it yet but i think the behaviour before was displaying the failed label if there was no known status so i added that fallback

ui/src/importer/importerTaskDetails/ImportedDocuments.js
NRodriguezcuellar: i see 🤔  i did not know that. Nor have i seen it doing weird things. I need it to run certain methods after the data prop has changed

ui/src/importer/importerTaskDetails/ImportedDocuments.js
NRodriguezcuellar: In most cases i just tried to preserve the condition logic that was in place because without having worked on it is hard to predict why that condition was placed there and what it might break.

ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImporterReportStatusLabel.js
zzacharo: Could you try to replicate the case in which the status is not present? It's worth checking the backend code to see if you can find a case that status is not returned.

NRodriguezcuellar: there are not that many cases actually. Have never seen it not being present

zzacharo: Then, I would say to remove it so that the code becomes more explicit! @ntarocco @kprzerwa wdyt?

NRodriguezcuellar: No hard opinion on this one 😄

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -1,194 +1,69 @@
...
-      () => this.checkForData(taskId),
-      invenioConfig.IMPORTER.fetchTaskStatusIntervalSecs
-    );
-    this.checkForData(taskId);
+    this.lastIndex = 0;
+    this.statistics = {
+      records: {
+        text: 'Records',
+        value: '',
+        filterFunction: record => record,
```
Pineirin, line 28: You are skipping using the filterFunction when you have all the records (which is right), so this line should be deleted ( I can do it in next PR so we merge ASAP).

Pineirin: Nevermind, I just saw you are using it to calculate the filteredRecords. You only are not using it in the statistics.

Pineirin (approved):
LGTM!

ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -287,97 +120,104 @@ export class ImportedDocuments extends React.Component {
...
   };
 
-  filteredRecords = () => {
-    const { statistics, data, selectedResult } = this.state;
-    const selectedFilterFunc = statistics[selectedResult].filterFunction;
+  filteredRecords = data => {
+    const { selectedResult } = this.state;
+    const selectedFilterFunc = this.statistics[selectedResult].filterFunction;
+
+    if (!data) return null;
```
ntarocco, line 127: It should return empty list?
```suggestion
    if (!data) return [];
```
Can we check this before calling it and remove the `if` from here? It looks weird to call this func if data is empty...

ntarocco (approved):

ui/src/importer/importerTaskDetails/ImportedDocuments.js
NRodriguezcuellar: That is not really possible because that if would block the entire render method. Now it will just show a loading screen for a quick moment. I returned the null because we have a check for that, even returning an empty array would give errors anyway.

kprzerwa (approved):

Issue #613, opened 2021-10-01
importer: error shows up only in the first line


kprzerwa:
Line 1: no error in preview: https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/123 but error on the e-item when importing: https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/124

task: investigate and fix the problem. Possible solution: not cleaned variable? not working label ? check the backend response

PR #650, opened 2021-11-02
Importer: displayed title from raw_json and added title check in addition to isbn check

NRodriguezcuellar:
The title from the parsed XML file is now being used instead of the title of the matched document. Now also checks if the title is equal to the matched document to further prevent wrongful matches 

closes https://github.com/CERNDocumentServer/cds-ils/issues/613

### DIFF
--- a/cds_ils/importer/documents/importer.py
+++ b/cds_ils/importer/documents/importer.py
@@ -208,9 +208,28 @@ class DocumentImporter(object):
             click.secho(e.original_exception.message, fg="red")
             db.session.rollback()
 
+    def validate_found_matches(self, not_validated_matches):
+        """Validate matched & parsed documents have same title/isbn pair."""
+        matches = []
+        partial_matches = []
+        document_class = current_app_ils.document_record_cls
+        title = self.json_data.get("title")
+
+        for pid in not_validated_matches:
+            document = document_class.get_record_by_pid(pid)
+            document_title = document['title']
+
+            if document_title.lower() == title.lower():
+                matches.append(document['pid'])
+            else:
+                partial_matches.append(document['pid'])
+
+        return matches, partial_matches
+
     def search_for_matching_documents(self):
         """Find matching documents."""
         document_class = current_app_ils.document_record_cls
+
         isbn_list = [
             identifier["value"]
             for identifier in self.json_data.get("identifiers", [])
@@ -230,7 +249,6 @@ class DocumentImporter(object):
             search = search_documents_by_isbn(isbn)
             results = search.scan()
             matches += [x.pid for x in results if x.pid not in matches]
-
         # check by doi
 
         for doi in doi_list:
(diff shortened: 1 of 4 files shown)


cds_ils/importer/documents/importer.py
```diff
@@ -198,9 +198,27 @@ def update_document(self, matched_document):
...
+        for pid in not_validated_matches:
+            document = document_class.get_record_by_pid(pid)
+            document_title = document['title']
+
+            if document_title == title:
+                matches.append(document['pid'])
+            else:
+                partial_matches.append(document['pid'])
+
+        return matches, partial_matches
```
kprzerwa, line 216: where are the partial matched displayed or consumed ?

NRodriguezcuellar: for now nowhere 😅

kprzerwa:
please edit the commit message to be compliant with the contributing guidelines

cds_ils/importer/documents/importer.py
```diff
@@ -266,7 +284,8 @@ def search_for_matching_documents(self):
                 document = document_class.get_record_by_pid(match)
                 if document.get("edition") != self.json_data.get("edition"):
                     matches.remove(match)
-        return matches
+
+        return self.validate_found_matches(matches)
```
kprzerwa, line 288: wouldn't be easier to use this function in the _match document instead of changing the returned type here ? 
what if you want to re-use this function without the validation ?

ui/src/importer/importerTaskDetails/ImportedDocumentReport.js
```diff
@@ -35,7 +35,7 @@ class ImportedDocumentReportComponent extends Component {
     if (!_isEmpty(documentReport.document)) {
       let volume = _get(documentReport, 'raw_json._serial[0].volume', '');
       volume = volume ? `(v. ${volume})` : '';
-      title = `${documentReport.document_json.title} ${volume}
+      title = `${documentReport.raw_json.title} ${volume}
```
kprzerwa, line 38: 👍

cds_ils/importer/documents/importer.py
NRodriguezcuellar: good one actually, i was hesitating where to apply it

kprzerwa:
I am not sure if you changed the commit message, but it is still too long (should be 55 chars)
fe.
```
Importer: change document report title source

* check matched title, identifiers pairs
```

cds_ils/importer/documents/importer.py
```diff
@@ -208,9 +208,28 @@ def update_document(self, matched_document):
...
+        matches = []
+        partial_matches = []
+        document_class = current_app_ils.document_record_cls
+        title = self.json_data.get("title")
+
+        for pid in not_validated_matches:
+            document = document_class.get_record_by_pid(pid)
+            document_title = document['title']
+
+            if document_title == title:
```
kprzerwa, line 222: there will be a problem with case sensitivity here, for this reason lowercase both titles for the comparison

kprzerwa (approved):

PR #654, opened 2021-11-05
[DEPENDS ON ILS] importer: match lowercase titles

### LINKED ISSUE #652: Importer - feedback 02/11/2021
**All sources**
- [x] Field 245, the indicators can be the following: 1st = 0 or 1 - 2nd = 0,1,2,3,4,5,6,7,8,9
See errors in https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/187
- [x] Field 100, the indicators can be the following: 1st = 0,1,3 - 2nd= empty or # 
- [x] The conversion of series should be done in the same way across the different sources: for EBL, the series title are created  without capital letters and for SPRINGER they are with capital letters. => Always keep the title as it is in the original file (but do the matching **ignoring** the case).  

**Springer import**
https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/184

- [x] BLOCKER When there is an existing  e-item (EBL or SAFARI), the importer should replace the existing item with the SPRINGER e-item, it doesn't seem to be the case anymore: ex. line 3 [978-981-15-0368-9], the existing EBL item is not deleted, and the SPRINGER item is not created. 
- [x] Lines 30 [978-3-030-00386-9]: 2 e-items seem to have been modified, but one is not working if we click on it: ex. https://sandbox-cds-ils.web.cern.ch/backoffice/eitems/mbwmz-sqb98 why is the problem here?
- [x] Line 42 [978-981-16-3568-7]: no e-item has been created, why? 
- [x] Line 6 [978-3-030-35709-2]: Error: "str: Multiple series found. " -> the importer indicates an error, however it seems the record has been created anyway, but without e-item. => the e-item should be created, and we can check the series manually.
- [x] The print ISBN (from 77680$z) seems correctly added for the updated records, but not for new records. Ex: Line 2 [978-3-030-26025-5]. 
- [x] BLOCKER The language "English" is still not added to the records, it should be added by default. 

**EBL import**
https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/181

- [x] When there is not defined role for the Author in the MARCXML (field 700 and field 100, $e), the role should be "Author" by default. (It works for SAFARI and SPRINGER).
- [x] BLOCKER There is still an issue with the matching of series, ex: https://sandbox-cds-ils.web.cern.ch/backoffice/series/dvt81-abk37 (existing series), and a new series has been created:  https://sandbox-cds-ils.web.cern.ch/backoffice/series/msdf5-krj06 : the only difference is the capital letters in one. The case SHOULD NOT be taken into account in the matching. 
- [x] Line 55 [EBC5927109]: Error: "str: Multiple series found. " -> the importer indicates an error, but it seems the record has been created anyway, but without e-item. => the e-item should be created, and we can check the series manually. 
 
**EBL delete**
https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/188

- [x] it seems the deletion doesn't work as it was working before: ID of records are not always indicated, see line 1 and 2. Why the id is not visible there? And the e-item is not deleted, nor the document. Why? This happens for many documents (110 on 627)
- [x] Line 22, it seems there were 2 e-items. It is strange there were 2 EBL e-items on the same document. We cannot check them anymore since they were deleted. Would it be possible to check what were those e-items, and if it is normal they were both deleted? 

**SAFARI delete**
https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/190

- [x] Many e-items, and documents not deleted: ex. line 5, the document exists in sandbox, but the e-item is not deleted, nor the document: https://sandbox-cds-ils.web.cern.ch/literature/kxts4-ys022 (same problem as mentioned for EBL)

kprzerwa:
* add proper eitem PIDs deletion
* fix duplicated series display
* close #652

### DIFF
--- a/cds_ils/importer/eitems/importer.py
+++ b/cds_ils/importer/eitems/importer.py
@@ -15,6 +15,7 @@ from invenio_app_ils.eitems.api import EItemIdProvider
 from invenio_app_ils.errors import IlsValidationError
 from invenio_app_ils.proxies import current_app_ils
 from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
 
 from cds_ils.importer.eitems.api import get_eitems_for_document_by_provider
 
@@ -27,15 +28,18 @@ class EItemImporter(object):
         json_metadata,
         eitem_json_data,
         metadata_provider,
-        current_provider_priority,
         provider_priority_sensitive,
         open_access,
         login_required,
     ):
         """Constructor."""
+        priority = current_app.config["CDS_ILS_IMPORTER_PROVIDERS"][
+            metadata_provider
+        ]["priority"]
+
         self.json_data = json_metadata
         self.metadata_provider = metadata_provider
-        self.current_provider_priority = current_provider_priority
+        self.current_provider_priority = priority
         self.is_provider_priority_sensitive = provider_priority_sensitive
         self.open_access = open_access
         self.login_required = login_required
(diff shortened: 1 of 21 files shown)


cds_ils/importer/eitems/importer.py
```diff
@@ -89,7 +93,17 @@ def _update_existing_record(self, existing_eitem, matched_document):
 
     def _delete_existing_record(self, existing_eitem):
         eitem_indexer = current_app_ils.eitem_indexer
-        existing_eitem.delete(force=True)
+        pid = existing_eitem.pid
```
ntarocco, line 96: Will this work after the deletion? It might be working only because you don't have a session commit yet.
I would probably, to be safe, save in vars:
```
pid_object_type, pid_object_uuid = pid.object_type, pid.object_uuid
```
same below...

ntarocco (approved):

cds_ils/importer/eitems/importer.py
kprzerwa: true, thanks for spotting it! it was working indeed because of the late commit, however in case of changes it is very easy to introduce bugs, I will fix it

PR #657, opened 2021-11-09
Import: nuanced errors on task page

NRodriguezcuellar:
Now shows the status code and error message when an error is returned on the task details page

Screenshot 2021-11-09 at 16 11 09


ui/src/importer/importerTaskDetails/ImportedDocuments.js
```diff
@@ -145,7 +145,12 @@ export class ImportedDocuments extends React.Component {
       data?.entries_count;
 
     if (error) {
-      return <NotFound title="Task not found" />;
+      const {
+        response: {
+          data: { message, status },
+        },
+      } = error;
+      return <NotFound title={status} message={message} isBackOffice />;
```
kprzerwa, line 153: isn't the name of the component misleading ? normally I would expect notFound = 404 only

ntarocco (commented):
Were you able to test/fix the double back issue when an error happens?

NRodriguezcuellar:
> Were you able to test/fix the double back issue when an error happens?

 tried to find in the code where it gets redirected but i could not find it and then got busy with my issues and PR's so set it aside.

kprzerwa (approved):

PR #658, opened 2021-11-10
importer: fix eitem search

### LINKED ISSUE #660: Wrong configuration for the Document field `unit_project`
The dynamic field `unit_project` is configured as string, but it should be a list instead.
When saving a migrated document that has this field set as a list, the web form shows an error because it expects a string.

Changes:
- Change the `config.py` to declare it as `List(SanitizedUnicode())`
- Change the `config.js` type to `type: 'array'`

kprzerwa:
* closes #660

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -455,7 +455,7 @@ ILS_RECORDS_METADATA_EXTENSIONS = {
         },
         "unit_project": {
             "elasticsearch": "keyword",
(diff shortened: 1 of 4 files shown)


jrcastro2 (approved):

ntarocco (approved):

Pineirin (approved):
:+1: I need to remember using the get in that cases.

Issue #499, opened 2021-04-30
Brief format Loan request - Add information about the available item


agentilb:
**Is your feature request related to a problem? Please describe.**

We need to display the `Copy location` and `shelf number` in the list of pending loans requests.
Otherwise, we need to open each pending request to be able to retrieve the documents. 
https://catalogue.library.cern/backoffice/loans?q=&f=state%3APENDING&f=availability%3AAvailable%20for%20loan&l=list&order=desc&p=1&s=15&sort=created

**Describe the solution you'd like**
Display `Copy location` and `shelf number` for the first available item, in Central Library, or in another Library, if not available in Central Library.

PR #676, opened 2021-11-25
Circulation: add to each loan the first item on shelf

Pineirin:
This change only affects elastic search.
The first item is the first of the associated document.
Allow removing fields from the loan serializer (configurable).
closes: https://github.com/CERNDocumentServer/cds-ils/issues/499
Screenshot from 2021-11-25 17-16-20

### DIFF
--- a/cds_ils/utils.py
+++ b/cds_ils/utils.py
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
             url["login_required_url"] = current_app.config[
                 "CDS_ILS_EZPROXY_URL"
             ].format(url=url["value"])
+
+
+def dynamic_delete_field(original_element, fields):
+    """Dynamically delete an array of fields from a dictionary."""
+    # Wrong parameter in the configuration
+    wrong_field = False
+
+    for field in fields:
+        if not isinstance(field, list):
+            original_element.pop(field)
+            continue
+        if not field:
+            continue
+
+        element = original_element
+        for subfield in field[:-1]:
+            if subfield in element.keys():
+                element = element[subfield]
+            else:
+                wrong_field = True
+                break
+
+        if wrong_field:
+            wrong_field = False
+            continue
+
+        element.pop(field[-1], None)
+
+    return original_element
--- a/cds_ils/circulation/indexer.py
+++ b/cds_ils/circulation/indexer.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Loan indexer APIs."""
+
+from invenio_app_ils.proxies import current_app_ils
+
+
+def index_extra_fields_for_loan(loan_dict):
+    """Indexer hook to modify the loan record dict before indexing.
+
+    The `on_shelf_item` fields are added to the loan only on the search index
+    because they are needed for search aggregation/filtering. They are not
+    needed when fetching the loan details.
+    """
+    document_class = current_app_ils.document_record_cls
+    document_record = document_class.get_record_by_pid(
+        loan_dict.get("document_pid")
+    )
+    document = document_record.replace_refs()
+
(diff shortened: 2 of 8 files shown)


cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
...
+                wrong_field = True
+                break
+
+        if wrong_field:
+            wrong_field = False
+            continue
+
+        element.pop(field[-1], None)
+
+    return original_element
```
Pineirin, line 47: I have gone a lot back and forth in this function.
I could do the same with three times less lines of code, but the current implementation provide a very robust handling to any inappropriate input in the configuration. 
The configuration of the fields passed to this function is documented in the config.py (# Format [field1, [field2, subfield1], [field3, subfield2, subsubfield1]]).

cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
...
+
+def dynamic_delete_field(original_element, fields):
+    """Dynamically delete an array of fields from a dictionary."""
+    # Wrong parameter in the configuration
+    wrong_field = False
+
+    for field in fields:
+        if not isinstance(field, list):
+            original_element.pop(field)
+            continue
```
Pineirin, line 29: This allows to define in the configuration [field1, [field2, subfield1]] instead of [[field1], [field2, subfield1]]

cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
...
+        if not field:
+            continue
```
Pineirin, line 31: This allows to define in the configuration [field1, []] (an empty array) without getting an error. An empty array should never be used regardless.

cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
...
+
+        element = original_element
+        for subfield in field[:-1]:
+            if subfield in element.keys():
+                element = element[subfield]
+            else:
+                wrong_field = True
+                break
```
Pineirin, line 39: Wrong field is used when you try to access of sub-field of an incorrect field, to skip the delete. Ej: if the configuration is [links [metodoto, pid]], we wouldn't get an error when why try to delete access pid in the non existing metodoto.

cds_ils/utils.py
kprzerwa: I think it is going a bit too deep, I would only choose to filter the fields in the "root" level for now

cds_ils/utils.py
kprzerwa: why would we define an empty list ? If the configuration is wrong it **should** throw an error

cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
             url["login_required_url"] = current_app.config[
                 "CDS_ILS_EZPROXY_URL"
             ].format(url=url["value"])
+
+
+def dynamic_delete_field(original_element, fields):
+    """Dynamically delete an array of fields from a dictionary."""
+    # Wrong parameter in the configuration
+    wrong_field = False
```
kprzerwa, line 24: what is it used for? what does it mean that a field is wrong?

cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
...
+
+    for field in fields:
+        if not isinstance(field, list):
+            original_element.pop(field)
+            continue
+        if not field:
+            continue
+
+        element = original_element
+        for subfield in field[:-1]:
```
kprzerwa, line 34: could you explain what is happening here?

cds_ils/utils.py
kprzerwa: I do not fully understand the concept, I think we should discuss in person

cds_ils/utils.py
```diff
@@ -16,3 +16,32 @@ def format_login_required_urls(urls):
...
+                element = element[subfield]
+            else:
+                wrong_field = True
+                break
+
+        if wrong_field:
+            wrong_field = False
+            continue
+
+        element.pop(field[-1], None)
```
kprzerwa, line 45: not sure why this is happening?

PR #680, opened 2021-12-01
Circulation: add to each loan the first item on shelf

Pineirin:
* This change only affects elastic search.
* The first item is the first of the associated document.
* Allow removing fields from the loan serializer.
* Closes CERNDocumentServer/cds-ils#499
* Improves: https://github.com/CERNDocumentServer/cds-ils/pull/676

### DIFF
--- a/cds_ils/circulation/indexer.py
+++ b/cds_ils/circulation/indexer.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# invenio-app-ils is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Loan indexer APIs."""
+
+from invenio_app_ils.proxies import current_app_ils
+
+
+def index_extra_fields_for_loan(loan_dict):
+    """Indexer hook to modify the loan record dict before indexing.
+
+    The `item_suggestion` fields are added to the loan only on the search index
+    because they are needed for search aggregation/filtering. They are not
+    needed when fetching the loan details.
+    """
+    document_class = current_app_ils.document_record_cls
+    document_record = document_class.get_record_by_pid(
+        loan_dict.get("document_pid")
+    )
+    document = document_record.replace_refs()
+
+    on_shelf = document.get("items", []).get("on_shelf", {})
+
+    if not on_shelf:
+        loan_dict["item_suggestion"] = {}
+        return
+
+    library = next(iter(on_shelf.values()))
+    item_suggestion = next(iter(library.values()))[0]
+
+    for field in [
+        "circulation_restriction",
+        "internal_location_pid",
+        "isbn",
+        "pid",
+    ]:
+        item_suggestion.pop(field, None)
+
+    loan_dict[
+        "item_suggestion"
+    ] = item_suggestion
--- a/cds_ils/circulation/serializers/__init__.py
+++ b/cds_ils/circulation/serializers/__init__.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Loan serializers."""
+from invenio_app_ils.circulation.serializers.csv import LoanCSVSerializer
+from invenio_app_ils.records.schemas.json import ILSRecordSchemaJSONV1
+from invenio_app_ils.records.serializers import record_responsify_no_etag
+from invenio_records_rest.serializers.response import search_responsify
+
+csv_v1 = LoanCSVSerializer(
+    ILSRecordSchemaJSONV1, csv_excluded_fields=["links"]
+)
+csv_v1_response = record_responsify_no_etag(csv_v1, "text/csv")
+csv_v1_search = search_responsify(csv_v1, "text/csv")
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -19,6 +19,8 @@ from datetime import timedelta
 
 from celery.schedules import crontab
 from invenio_app.config import APP_DEFAULT_SECURE_HEADERS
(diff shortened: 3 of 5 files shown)


cds_ils/circulation/indexer.py
```diff
@@ -0,0 +1,46 @@
...
+
+    library = next(iter(on_shelf.values()))
+    item_suggestion = next(iter(library.values()))[0]
+
+    for field in [
+        "circulation_restriction",
+        "internal_location_pid",
+        "isbn",
+        "pid",
+        "status",
```
kprzerwa, line 40: I think we should keep the status as the CSV file is being used for cataloging/stock management. Unless we have only CAN_CIRCULATE items indexed here, then indeed it can be removed

cds_ils/circulation/indexer.py
```diff
@@ -0,0 +1,46 @@
...
+    document_class = current_app_ils.document_record_cls
+    document_record = document_class.get_record_by_pid(
+        loan_dict.get("document_pid")
+    )
+    document = document_record.replace_refs()
+
+    on_shelf = document.get("items", []).get("on_shelf", {})
+
+    if not on_shelf:
+        loan_dict["on_shelf_item"] = {}
```
kprzerwa, line 29: I remember we agreed on naming the field "item_suggestion". Did anything change since ?

cds_ils/circulation/serializers/csv.py
```diff
@@ -0,0 +1,34 @@
...
+from invenio_app_ils.circulation.serializers.csv import \
+    LoanCSVSerializer as IlsLoanCSVSerializer
+
+
+class LoanCSVSerializer(IlsLoanCSVSerializer):
+    """Serialize loans."""
+
+    def __init__(self, *args, **kwargs):
+        """Initialization."""
+        kwargs['csv_excluded_fields'] = ["links"]
```
ntarocco, line 19: You don't need this. You can simply do the above, see other comment.

cds_ils/circulation/serializers/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+"""Loan serializers."""
+
+from invenio_app_ils.records.schemas.json import ILSRecordSchemaJSONV1
+from invenio_app_ils.records.serializers import record_responsify_no_etag
+from invenio_records_rest.serializers.response import search_responsify
+
+from .csv import LoanCSVSerializer
+from .json import LoanJSONSerializer
+
+csv_v1 = LoanCSVSerializer(ILSRecordSchemaJSONV1)
```
ntarocco, line 17:
```suggestion
from invenio_app_ils.circulation.serializers.csv import LoanCSVSerializer

csv_v1 = LoanCSVSerializer(ILSRecordSchemaJSONV1, csv_excluded_fields=["links"])
```

cds_ils/circulation/serializers/json.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
ntarocco, line 1: This class is probably not needed

cds_ils/config.py
```diff
@@ -411,6 +414,17 @@ def _parse_env_bool(var_name, default=None):
     "application/json": "cds_ils.series.serializers:json_v1_search",
     "text/csv": "cds_ils.series.serializers:csv_v1_search",
 }
+ILS_CIRCULATION_RECORDS_REST_ENDPOINTS[CIRCULATION_LOAN_PID_TYPE][
+    "record_serializers"
```
ntarocco, line 418: You can remove this one, no need to override the JSON serializer.

cds_ils/config.py
```diff
@@ -411,6 +414,17 @@ def _parse_env_bool(var_name, default=None):
...
+] = {
+    "application/json": "cds_ils.circulation.serializers:json_v1_response",
+}
+ILS_CIRCULATION_RECORDS_REST_ENDPOINTS[CIRCULATION_LOAN_PID_TYPE][
+    "search_serializers"
+] = {
+    "application/json": "cds_ils.circulation.serializers:json_v1_search",
```
ntarocco, line 425: Why not simply:
`ILS_CIRCULATION_RECORDS_REST_ENDPOINTS[CIRCULATION_LOAN_PID_TYPE]["search_serializers"]["text/csv"] = "cds_ils.circulation.serializers:csv_v1_search"`
So you don't need to touch the others.

cds_ils/circulation/indexer.py
```diff
@@ -0,0 +1,46 @@
...
+
+    The `on_shelf_item` fields are added to the loan only on the search index
+    because they are needed for search aggregation/filtering. They are not
+    needed when fetching the loan details.
+    """
+    document_class = current_app_ils.document_record_cls
+    document_record = document_class.get_record_by_pid(
+        loan_dict.get("document_pid")
+    )
+    document = document_record.replace_refs()
```
ntarocco, line 24: This is a really expensive operation. Was this tested when exporting a large number of loans?

cds_ils/circulation/indexer.py
```diff
@@ -0,0 +1,46 @@
...
+
+"""Loan indexer APIs."""
+
+from invenio_app_ils.proxies import current_app_ils
+
+
+def index_extra_fields_for_loan(loan_dict):
+    """Indexer hook to modify the loan record dict before indexing.
+
+    The `on_shelf_item` fields are added to the loan only on the search index
```
ntarocco, line 16: Docstring to be updated?

cds_ils/circulation/indexer.py
Pineirin: This code is called when we perform and indexing of the loans, not when we download the csv.

cds_ils/circulation/serializers/json.py
Pineirin: The json.py serializer was created because originally the process of deleting unnecessary fields was also used on the JSONs, as the algorithm I wrote could be easily reused there. In the implementation in invenio-records-rest, there is only the option of ignoring fields in the csv serializer. Originally the objective was only to delete fields on the csv serializer so, in the end, the json.py class is truly not needed.

PR #683, opened 2021-12-06
Scripts: add nginx certs

NRodriguezcuellar:


scripts/server
```diff
@@ -14,5 +14,5 @@ export FLASK_ENV=development
 export SERVER_NAME=127.0.0.1:5000
 
 invenio run \
-  --cert "$script_path"/../docker/nginx/test.crt \
-  --key "$script_path"/../docker/nginx/test.key
+  --cert "$script_path"/nginx/test.crt \
```
ntarocco, line 17: we are not using `nginx` here... I would propose to change the name of the folder simply to `certs`.

zzacharo (approved):

Issue #686, opened 2021-12-09
ElasticSearch fuzzy search bug.


Pineirin:
Testing in sandbox displayed the error: https://cds-sentry.web.cern.ch/sentry/cds-ils/issues/28448/?environment=sandbox&referrer=alert_email.
The error basically is a bug in the import process, which crashes with certain inputs.
As far as tested, the problem is caused by records with long titles and special characters.

PR #687, opened 2021-12-09
Importer: add an exception if fuzzy search breaks

Pineirin:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/686
Screenshot from 2021-12-14 13-35-20

### DIFF
--- a/cds_ils/importer/importer.py
+++ b/cds_ils/importer/importer.py
@@ -9,6 +9,7 @@
 import time
 
 import pkg_resources
+from elasticsearch.exceptions import ElasticsearchException
 from flask import current_app
 from invenio_app_ils.errors import RecordHasReferencesError
 from invenio_app_ils.proxies import current_app_ils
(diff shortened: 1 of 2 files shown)


kprzerwa:
could you provide an explanation for why you decided to go with this solution? Maybe ES documentation link, as discussed in the chat?

cds_ils/importer/importer.py
```diff
@@ -118,15 +119,23 @@ def _match_document(self):
 
         # fuzzy = trying to match similar titles and authors to spot typos
         fuzzy_results = self.document_importer.fuzzy_match_documents()
-        self.fuzzy_matches = [x.pid for x in fuzzy_results]
+        try:
+            self.fuzzy_matches = [x.pid for x in fuzzy_results]
+        except ElasticsearchException:
+            self.fuzzy_matches = None
```
kprzerwa, line 125: If you remember our conversation, I would simply raise an exception here, fe. `FuzzyMatchUnavailableException`
with a message "Due to technical limitation similarity match is unavailable for this title"
Thanks to that, you will simply display this error in the error column of the import report, and you don't need to change the fuzzy matches dictionary below. You just need to find the place where the exceptions are handled to add the `FuzzyMatchUnavailableException` to this place

Pineirin: I had a reason to avoid raising an exception, I will try to explain as best as I can.

In the importer, if we don't find a document with the exact same title during an import process, we will perform a fuzzy search. When we raise an exception in the importer, we catch that exception in the cds_ils/importer/api.py file and then we create a failure (in other words, we append to the list of records imported a failed one). 

In case there is a problem in the fuzzy search we shouldn't create a failed record import, as the record import could be successful regardless. One clear example would be the first import of a record. We would search for it with a normal ES search, without finding it, then we would perform a fuzzy search, in which we would get this "too_complex_to_determinize_exception". If in this case we raised an exception, that record wouldn't be imported when in fact it could be imported without any problem, even if the title is too complex for a fuzzy search in ES, it wouldn't be complex enough to suppose a problem if we added it to both ES and the DB.

IMO, is better to let the importer import records that cannot be partially matched and just display a message warning the user that we couldn't find partial matches for that specific record (because it's title is too complex).

kprzerwa:
concluded in the discussion that the issue requires bigger refactoring: error handling +  how the class attributes are set/get

Issue #526, opened 2021-05-31
Detailed record for documents (modifications required)


agentilb:
Here are the modification needed on the detailed view for documents

### Access online

- [x] When there is no ebook version of the document: below "Request the e-book", add a text: "We will do our best to give you access to the ebook, but please note that not all books are made available as ebooks by providers."

### Tab: "Detail": 

- [x] Do not display the source of the keywords. 

<img width="1059" alt="Screenshot 2021-05-31 at 11 54 47" src="https://user-images.githubusercontent.com/12181904/120192451-794e1d00-c21b-11eb-8f40-a56151bad67b.png">

- [x] The alternative titles are currently displayed behind "Others", they should be displayed at this level: 
-> the subtitle should be displayed beside the title, with a ":" to make the separation. 
-> The other alternative titles should be displayed below the main title having on the same line titles and subtitles. 

### Tab: "Identifier": 

- [x] Make identifiers clickable: Arxiv number, Inspire ID, DOI: 
<img width="1043" alt="Screenshot 2021-05-31 at 12 06 30" src="https://user-images.githubusercontent.com/12181904/120192496-8539df00-c21b-11eb-97cd-8578dd04d659.png">
<img width="1035" alt="Screenshot 2021-05-31 at 12 07 39" src="https://user-images.githubusercontent.com/12181904/120192510-89fe9300-c21b-11eb-95db-012d1cecc57c.png">

1334883 -> https://inspirehep.net/literature/1334883
arXiv:1412.5701 -> https://arxiv.org/abs/1412.5701
10.1093/oso/9780198856559.001.0001 -> doi.org/10.1093/oso/9780198856559.001.0001

### Tab: "Publications":

- [x] Replace "Publications" by "Publishing in" 
- [x] Show the Publication year (publication_info.year)
- [x] Display the relation to a periodical currently visible at the top of the details page in this tab, replacing the text "This is part of the series:" -> "This is part of the periodical". 

### Tab: "Conferences":

- [ ] #692
- [x] Do not display the CERN conference code 
- [x] "Series" -> "Series number"

### Tab: "Copyright":

- [x] License, the link to the license should be activated. 
- [x] Add a space between license material and "maintained by Creative Commons. 
- [x] Copyright: if not copyright notice, do not display parenthesis. 
[...]

PR #689, opened 2021-12-10
[DEPENDS]Vocab/doc request: added message and fields

NRodriguezcuellar:
Added a extra clarification message on the document request from in the front-site and also added some vocabulary fields to identifiers to make them clickable.

DESKTOP -> 

image

image

MOBILE -> 

Capture d’écran 2022-01-11 à 17 16 14

Capture d’écran 2022-01-11 à 16 32 58

closes https://github.com/CERNDocumentServer/cds-ils/issues/526
depends on https://github.com/inveniosoftware/react-invenio-app-ils/pull/515

### DIFF
--- a/cds_ils/vocabularies/data/alternative_identifier_schemes.json
+++ b/cds_ils/vocabularies/data/alternative_identifier_schemes.json
@@ -2,7 +2,8 @@
   {
     "type": "alternative_identifier_scheme",
     "key": "ARXIV",
-    "text": "arXiv"
+    "text": "arXiv",
+    "data": { "url_path":  "https://arxiv.org/abs/{identifier}"}
   },
   {
     "type": "alternative_identifier_scheme",
@@ -17,7 +18,10 @@
   {
     "type": "alternative_identifier_scheme",
     "key": "INSPIRE",
-    "text": "INSPIRE"
+    "text": "INSPIRE",
+    "data": {
+      "url_path": " https://inspirehep.net/literature/{identifier}"
+    }
   },
   {
     "type": "alternative_identifier_scheme",
--- a/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
+++ b/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
@@ -1,0 +1,11 @@
+import _isEmpty from 'lodash/isEmpty';
+
+export const LiteratureKeyword = ({ keywords }) => {
+  // hides the keyword sources
+  const keywordsIsEmpty = _isEmpty(keywords);
+
+  if (keywordsIsEmpty) return null;
+
+  const keywordValues = keywords.map(keyword => keyword?.value);
+  return keywordValues.join(';  ');
+};
--- a/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers/Identifiers.js
+++ b/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers/Identifiers.js
@@ -1,0 +1,153 @@
+import {
+  InfoPopup,
+  SeparatedList,
+} from '@inveniosoftware/react-invenio-app-ils';
+import capitalize from 'lodash/capitalize';
+import _isEmpty from 'lodash/isEmpty';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Divider, Table, Loader } from 'semantic-ui-react';
+
+// component not to be used in search pages
+// might trigger a lot of requests
+export class Identifiers extends Component {
+  componentDidMount() {
+    const { fetchVocabularyIdentifiers } = this.props;
+    fetchVocabularyIdentifiers();
+  }
+
+  render() {
+    const { identifiers, loading, vocabularyIdentifiers } = this.props;
+
+    const noIdentifiers = _isEmpty(identifiers);
+
+    if (noIdentifiers) {
+      return (
+        <>
+          <Divider horizontal>Identifiers</Divider>
+          'No identifiers'
+        </>
+      );
+    }
+
+    return (
+      <>
+        <Divider horizontal>Identifiers</Divider>
+        {loading ? (
+          <Loader />
+        ) : (
+          <Table definition>
+            <Table.Body>
+              <IdentifierRows
+                identifiers={identifiers}
+                vocabularyIdentifiers={vocabularyIdentifiers}
+              />
+            </Table.Body>
+          </Table>
+        )}
+      </>
+    );
+  }
+}
+
+Identifiers.propTypes = {
+  identifiers: PropTypes.array,
+  vocabularyIdentifiers: PropTypes.array,
+  fetchVocabularyIdentifiers: PropTypes.func.isRequired,
+  loading: PropTypes.bool.isRequired,
+};
+
+Identifiers.defaultProps = {
+  identifiers: [],
+  vocabularyIdentifiers: null,
+};
+
+export const IdentifierRows = ({
+  includeSchemes,
+  identifiers,
+  vocabularyIdentifiers,
+}) => {
+  const idsByScheme = {};
+
+  for (const id of identifiers) {
+    // Only include whitelisted schemes if includeSchemes is set
+    if (includeSchemes.length === 0 || includeSchemes.includes(id.scheme)) {
+      const value = createValueObject(
+        vocabularyIdentifiers,
+        id.value,
+        id.scheme,
+        id.material
+      );
+
+      if (id.scheme in idsByScheme) {
+        idsByScheme[id.scheme].push(value);
+      } else {
+        idsByScheme[id.scheme] = [value];
+      }
+    }
+  }
+
+  return Object.entries(idsByScheme).map(([scheme, ids]) => {
+    const values = ids.map(id => (
+      <>
+        {id.urlPath ? (
+          <a href={id.urlPath} target="_blank" rel="noreferrer noopener">
+            {id.value}
+          </a>
+        ) : (
+          id.value
+        )}
+
+        {id.material && (
+          <>
+            {' '}
+            <InfoPopup message="Material for this identifier">
+              ({capitalize(id.material)})
+            </InfoPopup>
+          </>
+        )}
+      </>
+    ));
+
+    return (
+      <Table.Row key={scheme}>
+        <Table.Cell width={4}>{scheme}</Table.Cell>
+        <Table.Cell>
+          <SeparatedList items={values} />
+        </Table.Cell>
+      </Table.Row>
+    );
+  });
+};
+
+const createValueObject = (vocabularyIdentifiers, value, scheme, material) => {
+  const noVocabulariesIdentifiers = !vocabularyIdentifiers?.length;
+  const defaultValue = { value, material };
+
+  if (noVocabulariesIdentifiers) return defaultValue;
+
+  const matchingVocabularyIdentifier = vocabularyIdentifiers.find(
+    identifier => identifier.value === scheme
+  );
+
+  const notValidForUrl =
+    !matchingVocabularyIdentifier || !matchingVocabularyIdentifier.url_path;
+
+  if (notValidForUrl) return defaultValue;
+
+  const urlPath = matchingVocabularyIdentifier.url_path.replace(
+    '{identifier}',
+    value
+  );
+
+  return { value, material, urlPath };
+};
+
+IdentifierRows.propTypes = {
+  includeSchemes: PropTypes.array,
+  identifiers: PropTypes.array,
+};
+
+IdentifierRows.defaultProps = {
+  includeSchemes: [],
+};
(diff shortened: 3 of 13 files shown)


cds_ils/vocabularies/data/alternative_identifier_schemes.json
```diff
@@ -17,7 +18,10 @@
   {
     "type": "alternative_identifier_scheme",
     "key": "INSPIRE",
-    "text": "INSPIRE"
+    "text": "INSPIRE",
+    "data": {
+      "url_path": " https://inspirehep.net/literature/{identifier}"
```
kprzerwa, line 23: we are missing a screenshot of how these links are displayed

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { Label } from 'semantic-ui-react';
```
kprzerwa, line 2: watch out for the linter warnings, unused values

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+export const LiteratureKeyword = ({ keywords }) => {
+  const keywordsIsValidArray = Array.isArray(keywords);
+
+  if (!keywordsIsValidArray) return null;
```
zzacharo, line 6: In which case the keywords are not an array?

zzacharo: Sorry, my question was more to challenge if keywords should be empty! Instead of checking here, I would probably ensure that the parent component is always providing an array! That way the component remains simple.

NRodriguezcuellar: ah i see! Well it is an overridden component that receives more general data so that it can be used flexibly. I would say the component is pretty simple either way :sweat_smile:

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
```diff
@@ -0,0 +1,211 @@
...
+  vocabularyApi,
+  withCancel,
+} from '@inveniosoftware/react-invenio-app-ils';
+import capitalize from 'lodash/capitalize';
+import _isEmpty from 'lodash/isEmpty';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Divider, Table, Loader } from 'semantic-ui-react';
+
+export class Identifiers extends Component {
```
ntarocco, line 13: This is not idea because if we want to display this link somewhere else, e.g. backoffice, we need to do the same thing, fetch from the backend for each vocab.
I am also not convinced by the fact that the UI should know that for this specific vocab, the URL exists and we need to fetch the vocab definition.
At the same time I don't have better ideas. 
ping @kprzerwa @zzacharo

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import _isEmpty from 'lodash/isEmpty';
+
+export const LiteratureKeyword = ({ keywords }) => {
```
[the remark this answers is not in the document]
kprzerwa: I thought it shouldn't since the request was to display less data (this is an override of existing component)

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
kprzerwa: the alternative is to hardcode the urls in the config like initially proposed, however it will require placing some conditional statements, also depending on the vocabularies, so it also creates similar dependency

cds_ils/vocabularies/data/alternative_identifier_schemes.json
```diff
@@ -2,7 +2,8 @@
   {
     "type": "alternative_identifier_scheme",
     "key": "ARXIV",
-    "text": "arXiv"
+    "text": "arXiv",
+    "data": { "url_path":  "https://arxiv.org/abs/{identifier}"}
```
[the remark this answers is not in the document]
kprzerwa: it was placed initially there, however I was concerned about the data dependency which you mention below, in case of differently defined vocabularies, the component dependent on this field will have issues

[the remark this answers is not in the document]
ntarocco: The UI should not depend on this. The UI should ideally check if, when rendering the vocab., the `data.url_path` exist and render it.
I am honestly not very convinced by fetching the URL from the vocab definition, too many REST calls.

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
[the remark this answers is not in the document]
ntarocco: ok it makes sense then. I would add a comment explaining why we are overriding it (displaying less data).

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
ntarocco: It does not: the dependency is a config, that can be changed.
The UI will simply use a generic config which, if set, will render the identifier as a link. There is no hardcoding in the code, it is config.

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculation/LiteratureKeyword.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
ntarocco, line 1: unused import?

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
```diff
@@ -0,0 +1,211 @@
...
+      data: { hits: alternativeIdentifierQuery },
+    } = await this.createQueryPromise('alternative_identifier_scheme');
+
+    const combinedHits = identifierQuery.concat(alternativeIdentifierQuery);
+
+    return combinedHits.map(this.serializer);
+  };
+
+  fetchVocabularies = async () => {
+    this.cancellableFetch = withCancel(this.query());
```
ntarocco, line 60: I have the feeling (maybe I am mistaken) that we are fetching each time... as discussed, it would be more efficient to fetch once and cache.

NRodriguezcuellar: Maybe that is a wrong interpretation on my part, how i understood it was that we did not want to make a separate call for each scheme and value, which does not happen. It does do 2 quick fetches when opening the tab itself, which in my opinion is really fast and light, so should be no problem. What do you think?

kprzerwa: why 2 fetches, not one ?

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
```diff
@@ -0,0 +1,211 @@
...
+import capitalize from 'lodash/capitalize';
+import _isEmpty from 'lodash/isEmpty';
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Divider, Table, Loader } from 'semantic-ui-react';
+
+export class Identifiers extends Component {
+  constructor(props) {
+    super(props);
+
```
kprzerwa, line 16: could you maybe add a disclaimer that this component shouldn't be used in search pages?

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
NRodriguezcuellar: because i have to fetch the identifiers + alternative identifiers

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
NRodriguezcuellar: sure :+1:

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
kprzerwa: I see, I forgot we had alternative identifiers there to. 
You could store it in redux so you could avoid the second request, but in this case I don't think it is crucial to do

ntarocco: I think it is not ok: if I go and visit multiple books, e.g. 10 books, for each of them there will be an HTTP call.
If you cache it in Redux instead, you have the HTTP calls only for the first book I visit.

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
NRodriguezcuellar: why exactly not again? 😅

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
NRodriguezcuellar: done!

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers.js
kprzerwa: because it might trigger many requests for vocabs one for each row of the search result
even with redux now, we will mount the instances of this component and some of them might perform multiple request before the state gets filled with vocab data

NRodriguezcuellar: ah ok! makes sense 😄

kprzerwa (approved):

ui/src/overridden/frontsite/Document/DocumentDetails/DocumentMetadataTabs/Identifiers/Identifiers.js
```diff
@@ -0,0 +1,154 @@
...
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import { Divider, Table, Loader } from 'semantic-ui-react';
+
+// component not to be used in search pages
+// might trigger a lot of requests
+export class Identifiers extends Component {
+  componentDidMount() {
+    const { vocabularyIdentifiers, fetchVocabularyIdentifiers } = this.props;
+    if (!vocabularyIdentifiers) {
```
ntarocco, line 16: I am sorry to stress on this, but here we are mixing responsibilities.
It is not a responsibility of a UI component to check for the existence of data and then do something: this code means that a UI component knows when and how an HTTP request has to be made. It also knows that `fetchVocabularyIdentifiers` is performing an HTTP call.
This should happen in the business logic layer:
* if you want to reuse `fetchVocabularyIdentifiers` in another cmp, you will have to reimplement the same `if`, over and over
* testing is harder

To be very clear: the UI should never know how the data is fetched.

Please refactor the code, here the pseudocode. 

Each specific vocab cmp should do:

```
componentDidMount(..)
  fetchVocabularyScheme(identifierScheme)
```

The Redux action, your business logic should hide the caching part to the external world:

```
  function fetchVocabularyScheme(identifierScheme)...
      identifierSchemeValue = mycache.getVocabularyScheme(identifierScheme)
      if not identifierSchemeValue:
          identifierSchemeValue = fetchFromBackend(identifierScheme)
          mycache.setVocabularyScheme(identifierSchemeValue)
      
      setState(... identifierSchemeValue ...)
```

Let's discuss this at the stand up.

setup.py
```diff
@@ -51,6 +51,7 @@
     "WTForms<3.0.0,>=2.3.3",
     # invenio-celery conflict fix
     "celery<5.2,>=5.1.0",
+    "tomli>=1.2.2, <2.0.0"
```
kprzerwa, line 54: let's chat about this

NRodriguezcuellar: sure!

kprzerwa: I meant a bit more detailed description, fe: `conflict caused by "black", waiting for a new version release`

NRodriguezcuellar: ah fair 😅 , will just take over your example then

ntarocco (approved):

PR #697, opened 2021-12-16
importer: improve the indexing after import

kprzerwa:
* address ConflictError on referenced records

### DIFF
--- a/cds_ils/importer/importer.py
+++ b/cds_ils/importer/importer.py
@@ -15,6 +15,7 @@ from invenio_app_ils.proxies import current_app_ils
 from invenio_app_ils.records_relations.api import RecordRelationsParentChild
 from invenio_app_ils.relations.api import Relation
 from invenio_db import db
+from invenio_indexer.api import RecordIndexer
 from invenio_pidstore.models import PersistentIdentifier
 from invenio_search import current_search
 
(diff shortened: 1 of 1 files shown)


cds_ils/importer/importer.py
```diff
@@ -148,13 +148,26 @@ def index_all_records(self):
...
-        eitem = self.eitem_importer.eitem_record
-        if eitem:
+        document_class = current_app_ils.document_record_cls
+        series_class = current_app_ils.series_record_cls
+        eitem_class = current_app_ils.eitem_record_cls
+
+        eitem_imported = self.eitem_importer.eitem_record
+        if eitem_imported:
+            # refetch record
+            eitem = eitem_class.get_record_by_pid(eitem_imported["pid"])
```
jrcastro2, line 158: Why do we need to refetch the record every time?

kprzerwa: to be sure we get the latest version, since the import is performed in the celery task

cds_ils/importer/importer.py
```diff
@@ -148,13 +148,26 @@ def index_all_records(self):
...
             eitem_indexer.index(eitem)
+            time.sleep(1)
```
ntarocco, line 160: why the `time.sleep`? :)

cds_ils/importer/importer.py
```diff
@@ -148,13 +148,26 @@ def index_all_records(self):
...
 
-        document_indexer.index(self.document)
+        # refetch record
+        document = document_class.get_record_by_pid(self.document["pid"])
+        document_indexer.index(document)
+        time.sleep(1)
```
ntarocco, line 165: Same here. If you need to wait for ES indexing, please use the config var ES DELAY as we are doing with the tasks.

ntarocco: And again below :)

zzacharo (approved):

PR #698, opened 2021-12-17
Documents: check if document exists when indexing

Pineirin:
Ensure application handles reindexing records related to a deleted document.
Issue appeared when reinxeding a document (and it's records) that was very recently deleted.
Add additional comments explaining rationale behind implementation of item_suggestion.

### DIFF
--- a/cds_ils/circulation/indexer.py
+++ b/cds_ils/circulation/indexer.py
@@ -8,6 +8,7 @@
 """Loan indexer APIs."""
 
 from invenio_app_ils.proxies import current_app_ils
+from invenio_pidstore.errors import PIDDeletedError
 
 
 def index_extra_fields_for_loan(loan_dict):
@@ -18,17 +19,26 @@ def index_extra_fields_for_loan(loan_dict):
     needed when fetching the loan details.
     """
     document_class = current_app_ils.document_record_cls
-    document_record = document_class.get_record_by_pid(
-        loan_dict.get("document_pid")
-    )
+    try:
+        document_record = document_class.get_record_by_pid(
+            loan_dict["document_pid"]
+        )
+    except PIDDeletedError:
+        # Document might have been deleted while reindexing asynchronously.
+        return
+
     document = document_record.replace_refs()
 
-    on_shelf = document.get("items", []).get("on_shelf", {})
+    on_shelf = document["items"]["on_shelf"]
 
+    # There may be no items on shelf, making impossible to suggest an item.
     if not on_shelf:
         loan_dict["item_suggestion"] = {}
         return
 
+    # Item_suggestion included due to a library request.
+    # The library wanted information about the first item available of each
+    # loan, in order to fasten the management of the pending requests.
     library = next(iter(on_shelf.values()))
     item_suggestion = next(iter(library.values()))[0]
 


kprzerwa (approved):
are there any other cases this could happen?

kprzerwa:
sorry, I find the commit message a bit confusing, because no error is raised in your code, so instead
```
Indexer: Handle errors when indexing a deleted document.
* ensure application raises error when a field that should be available isn't.
```

I would propose 
```
documents: check if document exists when indexing
```

cds_ils/circulation/indexer.py
```diff
@@ -18,9 +19,14 @@ def index_extra_fields_for_loan(loan_dict):
     needed when fetching the loan details.
     """
     document_class = current_app_ils.document_record_cls
-    document_record = document_class.get_record_by_pid(
-        loan_dict.get("document_pid")
-    )
+    try:
+        document_record = document_class.get_record_by_pid(
```
ntarocco, line 23: @Pineirin I didn't double-check, but from what I recall, I think that we saw that it might happen just after the `get_loans`, when checking for referenced relations?

cds_ils/circulation/indexer.py
```diff
@@ -18,12 +19,17 @@ def index_extra_fields_for_loan(loan_dict):
...
+            loan_dict["document_pid"]
+        )
+    except PIDDeletedError:
+        # Document might have been deleted while reindexing asynchronously.
+        return
+
     document = document_record.replace_refs()
 
-    on_shelf = document.get("items", []).get("on_shelf", {})
+    on_shelf = document["items"]["on_shelf"]
```
ntarocco, line 32: I guess that the extra below if now redundant... ping @kprzerwa to double check

ntarocco (approved):
LGMT

cds_ils/circulation/indexer.py
kprzerwa: I think the if below saves us some minimal dictionary operations and it could stay. Especially because we have cases of this field being empty

but after considering this part, I have some doubts if either solution is a good practice in the context of what we do below and maybe we should copy this dict instead of assigning value. Some lines below, we pop some values while still pointing at the original record. While here it will probably have no consequences, since it is indexer of a loan, I am worried it could happen in other places where there is a real possibility of updating the record dict. What do you think @ntarocco ?

kprzerwa: I discussed with @ntarocco , @Pineirin could you add a comment below, before line 38 explaining the reasoning of the following operations ?

ntarocco: To be 100% clear, I would add a comment explaining what `item_suggestion` is doing and why we are doing all those IFs...

cds_ils/circulation/indexer.py
```diff
@@ -18,17 +19,25 @@ def index_extra_fields_for_loan(loan_dict):
...
 
+    # There may be no items on shelf, making impossible to suggest an item.
     if not on_shelf:
         loan_dict["item_suggestion"] = {}
         return
 
+    # The item_suggestion field is obtained by selecting the first library that
```
kprzerwa, line 39: I meant more to describe the reason why we are doing it (adding the item_suggestion) in the first place

PR #701, opened 2021-12-17
importer: improve error handling

kprzerwa:
* refactor error handling


jrcastro2 (commented):
Why do we remove the `field` attribute when raising the `UnexpectedValue` exception?

ntarocco (approved):

kprzerwa:
> Why do we remove the `field` attribute when raising the `UnexpectedValue` exception?

@jrcastro2 because we pass it to the handler (as key), see
https://github.com/CERNDocumentServer/cds-ils/pull/701/files#diff-3bac5c3ad52745e6f89861d4fe3fd6fd0c303ae4a57276ccf5d5aeeeffedd48fR14
thanks to that there is no need to pass this parameter, the responsibility for creating the correct error message is now delegated to the handler

PR #707, opened 2021-12-21
Importer refactoring

kprzerwa:


cds_ils/importer/importer.py
```diff
@@ -154,82 +117,118 @@ def index_all_records(self):
...
+
         self._validate_provider()
-        self.action = "none"
-        # finds the exact match, update records
-        self.document = self._match_document()
 
-        if self.document:
-            self.output_pid = self.document["pid"]
-            self.delete_records(self.document)
+        exact_match, partial_matches = self._match_document()
```
ntarocco, line 204: In a future refactoring, this is also risky, same assumption that something happened before and maybe this `_match_document` will also set/change something in `self`.

cds_ils/importer/importer.py
```diff
@@ -346,7 +363,7 @@ def preview_delete_document(self, document):
         related_refs = set()
         for _, related_objects in document.relations.items():
             for obj in related_objects:
-                if not obj["record_metadata"].get("mode_of_issuance")\
+                if not obj["record_metadata"].get("mode_of_issuance") \
```
ntarocco, line 366: is it ok to use `.get` here? can it be unset?

ntarocco (commented):
It looks much better now ;)

ntarocco (approved):

cds_ils/importer/importer.py
kprzerwa: there are other related record types which don't have the mode of issuance

PR #715, opened 2022-01-10
ldap: delete not existing users

kprzerwa:


cds_ils/ldap/api.py
```diff
@@ -254,14 +252,18 @@ def import_new_ldap_users(new_ldap_users, log_func):
     )
 
 
-def delete_users(dry_run=True):
+def delete_users(dry_run=True, mark_for_deletion=False):
```
ntarocco, line 255: Should it be `True` by default maybe? It looks like that the `False` value is an admin thingy?

cds_ils/ldap/api.py
```diff
@@ -270,26 +272,43 @@ def delete_users(dry_run=True):
...
-            user_id = invenio_user.user_id
-            user = User.query.filter_by(id=user_id).one()
+            user_id = remote_account.user_id
+            success = False
+            ready_to_delete = True
+
+            if mark_for_deletion:
+                ready_to_delete = False
+                checks = invenio_user.mark_for_deletion()
+                db.session.commit()
```
ntarocco, line 296: I have the feeling that the `db.session.commit` should go inside the methods `mark/unmark`, WDYT?

cds_ils/ldap/api.py
```diff
@@ -270,26 +272,43 @@ def delete_users(dry_run=True):
     # get all Invenio remote accounts and prepare a list with needed info
     invenio_users = remap_invenio_users(log_func)
 
-    for invenio_user in invenio_users:
+    for remote_account in invenio_users:
```
ntarocco, line 275:
```suggestion
    for remote_account in remote_accounts:
```

cds_ils/ldap/api.py
```diff
@@ -270,26 +272,43 @@ def delete_users(dry_run=True):
...
+        try:
+            invenio_user = InvenioUser(remote_account)
+        except NoResultFound:
+            continue
+
         ldap_user = ldap_users_map.get(
-            invenio_user["remote_account_person_id"]
-        )
+                invenio_user.data["remote_account_person_id"]
```
ntarocco, line 282: Oh wow, did we miss this `data`?

cds_ils/ldap/tasks.py
```diff
@@ -32,3 +32,20 @@ def synchronize_users_task():
         db.session.rollback()
         current_app.logger.exception(e)
         log.set_failed(e)
+
+
+@shared_task
+def anonymize_users_task():
```
ntarocco, line 38: Question: I think that at the beginning, we added the `synchronize_users` method call to avoid a double fetch of users from LDAP.
Shall we do the same to be more efficient, instead of having a different celery task?

cds_ils/ldap/api.py
kprzerwa: that was my first impression too, but I followed in the end the other methods for consistency in the code. I will move it.

cds_ils/ldap/api.py
kprzerwa: 👍

cds_ils/ldap/api.py
kprzerwa: it was due to refactoring in the sync itself, and this method was overlooked in the refactoring, since it was not used

cds_ils/ldap/api.py
kprzerwa: I left false by default since it is used in cli as well, but I can replace it there too, and have `True` by default, it will be safer indeed

cds_ils/ldap/tasks.py
kprzerwa: I was thinking having separate task would be better in terms of separation of concerns. Especially if one of the tasks will be blocked (f.e. we cannot delete for some ldap related reasons), the import of new accounts would be blocked. What do you think ?

zzacharo: I like the separation of concerns in the 2 tasks and I agree with @kprzerwa in the non-blocking of the 2 tasks, anonymize and import of users. I think that if fetching twice the ldap users is a problem then we need to find another solution to cache the fetched ldap users and used them across our tasks. But this is something we need to do only if it's causing trouble I think.

ntarocco: Good point! 👏🏼

ntarocco (approved):

ntarocco:
LGTM! Tests are failing ;)

Issue #685, opened 2021-12-07
Issues found in sandbox in version v1.2.11


Pineirin:
- [x] For the SAF and EBL, there is only one thing missing: the rule should be: when there is already a SAF e-item, the EBL e-item should be added, when there is already an EBL e-item, the SAF e-item should be added. 
- [x] If the partial matches text is very big, the import table displays badly. https://sandbox-cds-ils.web.cern.ch/backoffice/importer/task/230
- [x] fix the math in the filters (see screen below)
- [ ] Very big imports are still impossible to visualize. (502 response code)
- [ ] add possibility to download the import xml file (if available, because they might be stored in `/tmp` - to be verified

<img width="2756" alt="Screenshot 2022-01-06 at 15 48 55" src="https://user-images.githubusercontent.com/38131488/148403444-63ae83c3-1faf-4655-bafe-a708419d9b44.png">

PR #720, opened 2022-01-14
Importer: Partial matches overflowing

NRodriguezcuellar:
partly closes https://github.com/CERNDocumentServer/cds-ils/issues/685


ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImportedDocumentReport.js
```diff
@@ -221,25 +229,31 @@ class ImportedDocumentReportComponent extends Component {
...
-                );
-              })}
-            </>
+            <Grid>
+              <Grid.Row>
+                <Grid.Column width={2}>
+                  <Icon color="red" name="exclamation" size="large" />
+                </Grid.Column>
+                <Grid.Column width={14}>
+                  <List>
```
kprzerwa, line 238: I think the list would be enough to push each element to a new row, adding the grid with one column is a bit overkill in my opinion. What do you think?

kprzerwa (approved):

PR #732, opened 2022-02-03
anonymization: improve email sending

### LINKED ISSUE #731: Anonymization: decrease n sent emails
When the anonymization cron runs, it tried to "delete" users when they have no loans/orders. If they have, it will send an email to the librarians to take actions.

The system should not send one email per user, but only one summary email at the end.

ntarocco:
* send only one summary email instead of one per user when the cron
  cannot anonymize user because they have loans/orders
* closes #731


run-tests.sh
```diff
@@ -19,6 +19,7 @@ function cleanup() {
 trap cleanup EXIT
 python -m check_manifest --ignore ".*-requirements.txt"
 python -m sphinx.cmd.build -qnNW docs docs/_build/html
+export POSTGRESQL_VERSION=12.4
```
ntarocco, line 22: What if we remove this?

tests/ldap/test_ldap.py
```diff
@@ -547,5 +522,65 @@ def _prepare():
...
+    mocker.patch("invenio_app_ils.patrons.anonymization.current_app_ils.patron_indexer.delete")  # noqa
+    # mock RemoteAccounts
+    mock1 = Mock()
+    mock1.user_id = 1  # patron 1
+    mock2 = Mock()
+    mock2.user_id = 2  # patron 2
+    mocker.patch("cds_ils.ldap.api.remap_invenio_users",
+                 return_value=[mock1, mock2])
+    # mock that the any `InvenioUser` exist
+    mocker.patch("cds_ils.ldap.api.InvenioUser")
```
ntarocco, line 558: is this working in the end?

ntarocco:
LGTM, thanks a lot!

tests/ldap/test_ldap.py
kpsherva: yes I changed the mock from where it was implemented to where it was used, the same with the mock below

PR #752, opened 2022-05-23
importer: match series by exact title

ntarocco:


cds_ils/importer/series/importer.py
```diff
@@ -231,32 +225,46 @@ def _validate_matches(self, json_series, matches):
...
-            for issn in json_series_issn:
-                if issn in existing_series_issn:
-                    issn_match = True
-            issns_not_equal = both_issns and not issn_match
+            # matching by exact title takes precedence over anything else
+            is_same_title = series["title"].lower().strip() == \
+                json_series_title
+            if is_same_title:
+                # title matching, stop and exit
+                return [match]
```
kpsherva, line 233: I think we are missing a case here where there are multiple series with the same title, it will match only first while it should fail with an exception. Unless matching only the first is OK

PR #765, opened 2022-08-24
importer: update Safari rules

ntarocco:


cds_ils/importer/providers/cds/helpers/parsers.py
```diff
@@ -208,13 +208,19 @@ def _clean(value_to_clean):
...
-        cleaned_values = []
-        for v in to_clean:
-            cleaned_values.append(_clean(v))
-        return cleaned_values
-    elif not multiple_values and type(to_clean) is tuple:
+    is_tuple = type(to_clean) is tuple
+    if is_tuple and not multiple_values:
         raise UnexpectedValue(subfield=subfield)
+
+    if multiple_values:
```
ntarocco, line 215: changed the return value to be always a `list`, and changed all usage of this func.

cds_ils/importer/eitems/importer.py
```diff
@@ -73,15 +79,19 @@ def _set_record_import_source(self, record_dict):
...
+        PROVIDERS = current_app.config["CDS_ILS_IMPORTER_PROVIDERS"]
+
+        eitem_provider = self._get_record_import_provider(eitem)
+        if eitem_provider:
+            provider_definition = PROVIDERS.get(eitem_provider, {})
+            eitem_priority = provider_definition.get(
+                "priority", PRIORITY_FOR_UNKNOWN
+            )
+        else:
+            eitem_priority = PRIORITY_FOR_UNKNOWN
```
kpsherva, line 94:
```suggestion
       eitem_priority = PRIORITY_FOR_UNKNOWN
       
       provider_definition = PROVIDERS.get(eitem_provider, {})
       eitem_priority = provider_definition.get(
          "priority", PRIORITY_FOR_UNKNOWN
      )
```

cds_ils/importer/eitems/importer.py
```diff
@@ -168,9 +178,9 @@ def _should_import_eitem_by_priority(self, matched_document):
 
         comparison_list = []
         for eitem in existing_eitems:
-            is_imported_or_created = self._is_imported(
-                eitem
-            ) or self._is_manually_created(eitem)
+            is_imported_or_created = self._is_imported(eitem) \
```
kpsherva, line 181: I think we can completely remove this check, since we have always imported, created manually or migrated, unless you want to leave it for "undefined" cases (which should not happen by the way, I think created_by fields were required)

cds_ils/importer/eitems/importer.py
```diff
@@ -371,9 +381,9 @@ def preview_import(self, matched_document):
                 )
                 for hit in document_eitems:
                     eitem = eitem_cls.get_record_by_pid(hit.pid)
-                    is_imported_or_created = self._is_imported(
-                        eitem
-                    ) or self._is_manually_created(eitem)
+                    is_imported_or_created = self._is_imported(eitem) \
+                        or self._is_manually_created(eitem) \
+                        or self._is_migrated(eitem)
```
kpsherva, line 386: same here

cds_ils/importer/providers/safari/rules/document.py
```diff
@@ -38,17 +38,17 @@ def agency_code(self, key, value):
     return value
 
 
-@model.over("authors", "(^1001)|(^7001)")
+@model.over("authors", "(^1001)|(^1000)|(^7001)")
```
kpsherva, line 41:
```suggestion
@model.over("authors", "(^100)|(^7001)")
```
will cover all the variations of ind1 for key starting with 100

cds_ils/importer/eitems/importer.py
ntarocco: Keep it for extra safety? :)

cds_ils/importer/eitems/importer.py
ntarocco: OK fixed, I was not super happy to potentially search in `PROVIDERS` for a `None` or empty value, but I am ok with that.

cds_ils/importer/providers/safari/rules/document.py
ntarocco: What if there is a `1002`? Anyway, fixed it.

kpsherva: it is the same field still, I think safari started to use the indicators internally for something, however the field is still author, which is what we want to cover

PR #791, opened 2022-11-10
importer: fix normalized title to ignore only last 'series'

ntarocco:


cds_ils/importer/series/importer.py
```diff
@@ -103,7 +104,13 @@ def _update_field_identifiers(self, matched_series, json_series):
     @staticmethod
     def _normalize_title(title):
         """Return a normalized title."""
-        return title.lower().replace(" series", "").replace(" Series", "").strip()
+        t = title.lower()
+        # remove `series` only at the end of the title
+        # `International Series of Numerical Mathematics series`
```
kpsherva, line 109: does it make sense adding this case to the tests?

ntarocco: I did add it to the tests. What do you mean exactly?

PR #797, opened 2023-01-05
importer: add default language for safari

ntarocco:


cds_ils/importer/providers/safari/safari.py
```diff
@@ -31,7 +31,10 @@ def do(
     ):
         """Overwrite the do method."""
         init_fields = deepcopy(self._default_fields)
-        return super().do(blob, ignore_missing, exception_handlers, init_fields)
+        mapped = super().do(blob, ignore_missing, exception_handlers, init_fields)
+        # ensure language as a default value, when not define in the Safari input
+        mapped.setdefault("languages", ["ENG"])
```
kpsherva, line 36: why not to add it in the default fields ? (line 23)
I find it a bit strange to manipulate the dictionary here, after all the transformations are done

ntarocco: Eh eh :) I have added there as first solution, but the issue is that the decorator [@for_each_value](https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/importer/providers/safari/rules/document.py#L177), will keep what is there and append.
What happens is that, then, the `ENG` will always be there, and then the languages parsed in the Safari input will be appended.

I am obliged to `setdefault` **after** the transformation. 
If you prefer, I can move ALL default fields after... not sure if I break anything.

kpsherva: it will break if you move them after, because if I remember correctly, they will overwrite the output from tranformation rules.
assigning the default value inside the rule: https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/importer/providers/safari/rules/document.py#L179 would be an option too, I would say it would be easier to find where the value is assigned, otherwise it can stay like this

kpsherva (approved):

Issue #729, opened 2022-01-26
Wording


salomerohr:
Would it be possible to change some words to improve comprehension:

Borrowing requests:
Due date -> Returned date

Purchase orders :
Internal purchase requisition ID (DAI)
ID of inter-departmental transfer (TID)
IDT ID -> TID ID
IPR ID -> DAI ID

Search bar:
ebooks -> e-books

Left menu:
Series / Monographs -> Series / Multipart

salomerohr:
e-book vs. ebook: 
https://catalogue.library.cern/backoffice/eitems?q=urls.description%3A%22ebook%22&l=list&order=asc&p=1&s=15&sort=bestmatch 
vs. 
https://catalogue.library.cern/backoffice/eitems?q=urls.description%3A%22e-book%22&l=list&order=asc&p=1&s=15&sort=bestmatch 
-> I thought that the chosen word was 'e-book' according to the form used by the Oxford Dictionary.

salomerohr:
harmonize ebook / e-book in e-items.

salomerohr:
On the form: https://sandbox-cds-ils.web.cern.ch/backoffice/acquisition/orders/create
Internal purchase requisition ID -> please add 'DAI' in parenthesis at the end
ID of inter-departmental transfer -> please add 'TID' in parenthesis at the end

I still see 'ebooks' in the Search bar in the front office:
ebooks -> e-books

I do not see this change. I still see 'Monographs':
Left menu:
Series / Monographs -> Series / Multipart

lydipi:
In addition to Salomé's comments, could you please change "Returned Date" with "Returned date":
<img width="1163" alt="Capture d’écran 2024-03-13 à 11 41 20" src="https://github.com/CERNDocumentServer/cds-ils/assets/148760346/960db55e-4c6b-4f5c-ae1a-7779fe272950">

PR #851, opened 2024-01-16
UI: overridden: Override PaymentInformation and OrderLine

sakshamarora1:
IPR -> DAI
IDT -> TID

Closes: #729


ui/src/overridden/backoffice/utils.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { config } from "../../config";
```
ntarocco, line 2: Check the warning above

ui/src/overridden/backoffice/utils.js
```diff
@@ -0,0 +1,20 @@
...
+
+export const formatPrice = (price, includeCurrency = true) => {
```
ntarocco, line 4: Why not re-using the already defined func [here](https://github.com/inveniosoftware/react-invenio-app-ils/blob/master/src/lib/api/utils.js#L69)?

ui/src/config.js
```diff
@@ -146,6 +146,43 @@ export const config = {
   },
   ILL_BORROWING_REQUESTS: {
     defaultType: "PHYSICAL_COPY",
+    fieldOverrides: {
```
kpsherva, line 149: how is this key used? I think I miss some context

kpsherva (approved):

Issue #820, opened 2023-10-23
Ebook importation: direct link for documents


lydipi:
In backoffice, e-books importer. Would it be possible to make every PID as a link? for example when I do an import, there is direct link to the concerned documents. But for deletions, there is no clickable link for document, just the PID. You can see an example here:
<img width="1263" alt="nolink_PID" src="https://github.com/CERNDocumentServer/cds-ils/assets/148760346/b70bbb7b-4814-4b57-b34c-9b94517c5299">

kpsherva:
possible solutions:
- [ ] adding the link before storing the message as an error (basically adding an `<a>` inside the error string
- [ ] parsing the error on the frontend to identify the PIDs - not ideal

PR #855, opened 2024-02-21
importer: Rename Output->Related Document and link doc for errored rows

sakshamarora1:
Closes: #820 

Screenshot 2024-02-21 at 11 55 35

### DIFF
--- a/cds_ils/importer/handlers.py
+++ b/cds_ils/importer/handlers.py
@@ -87,8 +87,11 @@ def importer_exception_handler(
     exc, log_id, record_recid, json_data=None, *args, **kwargs
 ):
     """Handle importer exception."""
+    report = {"raw_json": json_data}
+    if hasattr(exc, "record_id"):
+        report["output_pid"] = exc.record_id
     ImportRecordLog.create_failure(
-        log_id, record_recid, str(exc.description), report={"raw_json": json_data}
+        log_id, record_recid, str(exc.description), report=report
     )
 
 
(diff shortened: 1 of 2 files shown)


ui/src/importer/importerTaskDetails/ImporterTaskDetailsComponents/ImportedTable.js
```diff
@@ -35,7 +35,7 @@ export class ImportedTable extends React.Component {
                 Action
               </Table.HeaderCell>
               <Table.HeaderCell collapsing rowSpan="3">
-                Output document
+                Related document
```
jrcastro2, line 38: Comment: If this change was discussed with the library then it's good to me. Otherwise be aware that this will affect to correclty processed records as well.
I would check with them before.

sakshamarora1: This was discussed! 😄

cds_ils/importer/handlers.py
```diff
@@ -87,8 +87,11 @@ def importer_exception_handler(
     exc, log_id, record_recid, json_data=None, *args, **kwargs
 ):
     """Handle importer exception."""
+    report = {"raw_json": json_data}
+    if hasattr(exc, "record_id"):
+        report["output_pid"] = exc.record_id
```
kpsherva, line 92:
```suggestion
    output_pid = getattr(exc, "record_id", None)
    report = {"raw_json": json_data, "output_pid": output_pid}
```

kpsherva (approved):

Issue #865, opened 2024-03-13
Importer: import audiobooks


kpsherva:
- An Audiobook can be of Document Type Book. Audiobook are added to existing Book record as additional e-item with description 'Audiobook'. If there is no record, a Book record is created with an e-item with description 'Audiobook'. ISBNs identified as '(electronic audio bk.)' (020 subfield q) should refer to Material 'Audiobook'. The other ISBNs should be treated as usual. Important: `<leader>` field should include value `im` to identify as audiobook
- Display Audiobook  in the Formats in the facets (based on the description in the e-items): e-book/ebook (to be standardized - please bulk update to 'e-book') -> Facet 'E-book'; Audiobook -> Facet 'Audiobook';  in the Filter by Format.
- For Deletions, distinguish Ebook/Audiobook with the LDR field (am (Book)/im (Audiobook)) to remove only the corresponding e-item. The rest of deletion process would as usual.

salomerohr:
I tested with 2 files and got the following results:
- 'videos' are well imported as such: example: https://sandbox-cds-ils.web.cern.ch/literature/08j5p-vsf95 Works!
- 'audiobook' that matched with an 'e-book': https://sandbox-cds-ils.web.cern.ch/backoffice/documents/y32s1-qzf62 it looks the audiobook item replaced the e-book (it should have been added - there should be 2 e-items (the 'e-book and the 'audiobook').
- 'audiobook' (without already existing record - new record): example: https://sandbox-cds-ils.web.cern.ch/backoffice/documents/e4k6y-c1235 Works!

You can ignore the following fields in the matching rules: {'758__4', '758__a', '758__i', '758__1'} in the importer.

salomerohr:
Please let us know when point 2 in the above comment is solved, and we will also test if the importer is behaving as it should for deletions (should delete the corresponding e-item (either e-book or audiobook) if both are present).

lydipi:
Perfect, it is working with the importer for audiobooks! it has added the audiobook without deleting the e-book.

salomerohr:
Ok!

PR #873, opened 2024-03-26
importer: Add ability to import audiobooks

sakshamarora1:
closes: #865

### DIFF
--- a/tests/importer/data/safari_audiobook.xml
+++ b/tests/importer/data/safari_audiobook.xml
@@ -1,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<collection xmlns="http://www.loc.gov/MARC21/slim" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.loc.gov/standards/marcxml/schema/MARC21slim.xsd">
+<record xmlns="http://www.loc.gov/MARC21/slim">
+    <leader>00000nim a22000007i 4500</leader>
+    <controlfield tag="001">on1417409648</controlfield>
+    <controlfield tag="003">OCoLC</controlfield>
+    <controlfield tag="005">20240125213018.0</controlfield>
+    <controlfield tag="006">m     o  h        </controlfield>
+    <controlfield tag="007">sz zunnnnnuneu</controlfield>
+    <controlfield tag="007">cr nnannnuuuuu</controlfield>
+    <controlfield tag="008">240117s2024    xx nnnn o      z  n eng d</controlfield>
+    <datafield tag="040" ind1=" " ind2=" ">
+      <subfield code="a">ORMDA</subfield>
+      <subfield code="b">eng</subfield>
+      <subfield code="e">rda</subfield>
+      <subfield code="e">pn</subfield>
+      <subfield code="c">ORMDA</subfield>
+      <subfield code="d">OCLCO</subfield>
+    </datafield>
+    <datafield tag="020" ind1=" " ind2=" ">
+      <subfield code="a">9781663731913</subfield>
+      <subfield code="q">(electronic audio bk.)</subfield>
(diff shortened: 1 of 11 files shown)


cds_ils/importer/providers/cds/rules/base.py
```diff
@@ -535,7 +535,9 @@ def isbns(self, key, value):
     for v in force_list(value):
         subfield_u = clean_val("u", v, str)
         isbn = {
-            "value": clean_val("a", v, str) or clean_val("z", v, str),
+            "value": clean_val("a", v, str)
+            or clean_val("z", v, str)
+            or clean_val("q", v, str),
```
kpsherva, line 540: where is the q coming from?

cds_ils/importer/providers/ebl/rules/document.py
```diff
@@ -125,6 +126,14 @@ def identifiers(self, key, value):
         }
         if isbn not in _identifiers:
             _identifiers.append(isbn)
+    if "q" in value and self.get("_eitem", {}).get("_type", None) == "audiobook":
```
kpsherva, line 129: this part is a bit unreadable, how about extracting
```
_eitem_type = _eitem.get("_type", "e-book")
```
on the top of the method and then use the variable here ?

cds_ils/importer/providers/ebl/rules/document.py
```diff
@@ -125,6 +129,14 @@ def identifiers(self, key, value):
         }
         if isbn not in _identifiers:
             _identifiers.append(isbn)
+    if "q" in value and _eitem_type == "audiobook":
+        isbn = {
+            "scheme": "ISBN",
+            "value": clean_val("q", value, str, req=True),
```
kpsherva, line 135: to be removed, discussed IRL: EBL will not have the audiobooks. also value of subfield q should not be stored in "value" but in "material"

cds_ils/importer/providers/ebl/ebl.py
```diff
@@ -30,8 +30,10 @@ def do(
         init_fields=None,
     ):
         """Overwrite the do method."""
-        init_fields = deepcopy(self._default_fields)
-        return super().do(blob, ignore_missing, exception_handlers, init_fields)
+        fields = deepcopy(self._default_fields)
```
kpsherva, line 33: to remove for now, EBL won't have audiobooks

cds_ils/importer/providers/springer/springer.py
```diff
@@ -30,8 +30,10 @@ def do(
         init_fields=None,
     ):
         """Overwrite the do method."""
-        init_fields = deepcopy(self._default_fields)
-        return super().do(blob, ignore_missing, exception_handlers, init_fields)
+        fields = deepcopy(self._default_fields)
```
kpsherva, line 33: to remove, springer will not have audiobooks

kpsherva:
just realised, aren't we still missing from this PR "material" or "type" field to be added on the e-item record?

kpsherva (approved):

tests/importer/data/safari_audiobook.xml
```diff
@@ -0,0 +1,139 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kpsherva, line 1: @ntarocco, @zzacharo  I had some second thoughts about exposing the below metadata in the tests, but since the e-book is behind login I think it is fine, WDYT?

zzacharo: what exactly are you worried about in terms of metadata exposure?

kpsherva: @zzacharo if we can use the book as an example in tests

zzacharo: I don't see any private data in the test record, but worst case you can alter some of the metadata.

PR #893, opened 2024-05-23
document: CERN Maps integration improvements

kpsherva:


ui/src/overridden/backoffice/Items/ItemCirculationShelf.js
```diff
@@ -1,26 +1,19 @@
...
-          rel="noreferrer"
-          color="blue"
-          fluid
-        >
-          <Icon name="map pin" />
-          SHELF {metadata.shelf}
-        </Button>
-      </>
+      <Icon name="map pin" />
+      Call number {callNumber ? callNumber.value : "missing"} (SHELF: {metadata.shelf})
```
sakshamarora1, line 16: Since shelf is not a mandatory field, we should handle empty shelf too, or else it would show `Call number XYZ.123 (SHELF: undefined)`

ui/src/overridden/backoffice/Items/ItemCirculationShelf.js
```diff
@@ -1,26 +1,20 @@
...
-import { Grid, Embed, Button, Icon } from "semantic-ui-react";
+import { Grid, Icon } from "semantic-ui-react";
 import PropTypes from "prop-types";
-import { shelfLink } from "../../utils";
 
 export const ItemCirculationShelf = ({ metadata }) => {
+  let callNumber = {};
+  if (!isEmpty(metadata.identifiers)) {
+    callNumber = metadata.identifiers.find(
+      (identifier) => identifier.scheme === "CALL_NUMBER"
```
ntarocco, line 10: I might have missed this in other PR, but what is the id scheme `CALL_NUMBER`?

ntarocco (commented):
LGTM! Tests are not happy :)

ui/src/overridden/backoffice/Items/ItemCirculationShelf.js
kpsherva: it is the old shelf "identifier" which we were asked to preserve and store as call number, since we now have the actual self numbers

salomerohr:
To test, I added a Shelf number to : https://sandbox-cds-ils.web.cern.ch/backoffice/items/nrv6q-1dr72 (selection in the list) but the Call number disappeared without me deleting it; and I do not see it in the form anymore. It looks like that now (missing Call number):
image

The rest looks fine!

salomerohr:
And Call number on the items is mandatory (and should not be deletable).

salomerohr:
Yes!

Issue #651, opened 2021-11-02
Distinguish First names and Last name for patrons


salomerohr:
It is difficult to distinguish where the First name finishes and the Last name starts for patrons, e.g.:
Joao Gentil Mendes Saraiva

It would be useful e.g. when processing loan requests to correctly sort the books for pick up at the Library desk.

**Describe the solution you'd like**
It is clearer in the phonebook as the last name is in capital letters:
Joao Gentil MENDES SARAIVA

**Describe alternatives you've considered**
Or, if easier, separate first names and last name with a comma:
Joao Gentil, Mendes Saraiva
or
Mendes Saraiva, Joao Gentil

PR #903, opened 2024-06-11
LDAP: Update user names formatting on sync

sakshamarora1:
closes: #651

### DIFF
--- a/cds_ils/oauth.py
+++ b/cds_ils/oauth.py
@@ -1,42 +1,13 @@
-from flask import session, current_app, flash, redirect, g
-from jwt import decode
+# Copyright (C) 2024 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
 
+"""CDS-ILS oauth overrides."""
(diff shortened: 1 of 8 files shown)


cds_ils/oauth.py
```diff
@@ -0,0 +1,92 @@
...
+
+    return dict(
+        user=dict(email=email.lower(), profile=dict(username=nice, full_name=name)),
+        external_id=external_id,
+        external_method="cern_openid",
+        active=True,
+    )
+
+
+def account_info(remote, resp):
```
kpsherva, line 69: don't we only need some of these functions? even better, can we have configurable callable to inject our own _account_info? it would require adding this variable in the oauthclient (you will need to make sure the default callable is the one defined in oauthclient)

tests/conftest.py
```diff
@@ -121,7 +121,7 @@ def patrons(app, db):
         **dict(
             user_id=user_id,
             _displayname="id_" + str(user_id),
-            full_name="System User",
+            full_name="USER, System",
```
kpsherva, line 124: I think this one we can leave as it is :D

kpsherva (approved):

Issue #529, opened 2021-05-31
Improve the search behaviour


agentilb:
**Is your feature request related to a problem? Please describe.**
Some users might be confused by the search behaviour, it is high visible for the standards (#516), but it creates confusion on other document types as well.  

2 aspects seem to be improved: 
- When searching 2 or more terms in the search, the implicite boolean should be AND and not OR (this is the case in Inspire)
- The order of the terms shouldn't have an impact on the search results, especially if they are part of the title and the author. 

Some examples are explained here: 
https://docs.google.com/document/d/1Xw4I4sOwHTNqKYR6JrhOfMkPwlNTuE4cmZgGKmbTzDE/edit#

salomerohr:
Another example where it should be an AND between the words:
authors.full_name:(Albert Einstein) - literature from a specific author

PR #908, opened 2024-06-19
config: use AND query parser instead of default operator OR

sakshamarora1:
closes: #529

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -59,6 +59,7 @@ from invenio_circulation.transitions.transitions import (
 from invenio_oauthclient.contrib import cern_openid
 from invenio_records_rest.schemas.fields import SanitizedUnicode
 from invenio_records_rest.utils import deny_all
+from invenio_search.engine import dsl
 from marshmallow.fields import Bool, List
 
 from .circulation.utils import circulation_cds_extension_max_count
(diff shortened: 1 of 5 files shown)


cds_ils/config.py
```diff
@@ -398,13 +405,15 @@ def _parse_env_bool(var_name, default=None):
...
 }
+RECORDS_REST_ENDPOINTS[LITERATURE_PID_TYPE]["search_query_parser"] = query_parser_and
 RECORDS_REST_ENDPOINTS[SERIES_PID_TYPE]["record_serializers"] = {
     "application/json": "cds_ils.series.serializers" ":json_v1_response"
 }
 RECORDS_REST_ENDPOINTS[SERIES_PID_TYPE]["search_serializers"] = {
     "application/json": "cds_ils.series.serializers:json_v1_search",
     "text/csv": "cds_ils.series.serializers:csv_v1_search",
 }
+# RECORDS_REST_ENDPOINTS[SERIES_PID_TYPE]["search_query_parser"] = query_parser_and
```
kpsherva, line 416: any reason for commenting out?

cds_ils/config.py
```diff
@@ -85,6 +86,11 @@ def _parse_env_bool(var_name, default=None):
     return default
 
 
+def query_parser_and(qstr, extra_params):
```
ntarocco, line 89: May I suggest to, instead of creating an entire query_parser func, simply create/use a hook that allows changing the extra params?

```
def query_params_modifier(query_params):
    query_params["default_operator"] = "AND"
    return query_params
```
basically a hook to allow modifying any query params.

WDYT?

cds_ils/config.py
sakshamarora1: Oops, corrected

PR #916, opened 2024-08-30
search: enable AND operator and cross_field

ntarocco:
Original issue:
- Book cannot be found: https://catalogue.library.cern/literature/0s8wv-bxz77
- Query: https://catalogue.library.cern/search?q=quantum%20field%20theory%20schwartz&l=grid&order=asc&p=1&s=15&sort=bestmatch
- NEEDS https://github.com/inveniosoftware/invenio-app-ils/pull/1221

### DIFF
--- a/cds_ils/config.py
+++ b/cds_ils/config.py
@@ -87,7 +87,24 @@ def _parse_env_bool(var_name, default=None):
 
 def query_params_modifier(extra_params):
     """Modifier for parameters to dsl Query function."""
+    # The `AND` operator, by default, will return hits where the query tokens are all
+    # in the same field. `cross_fields` must be applied to enable cross fields searches.
     extra_params["default_operator"] = "AND"
+    # The `cross_fields` search type requires all searchable fields to have the same
+    # analyzers. In the example below, `title` and `authors.full_name` must have the
+    # same analyzer defined in the mappings.
+    #
+    # "query_string": {
+    #   "query": "Theory Schwartz",
+    #   "type": "cross_fields",
+    #   "fields": ["title", "authors.full_name", "*"]
+    #   "analyzer": "custom_analyzer"
+    # }
+    #
+    # Explanation:
+    # "((+title:theory +authors.full_name:theory) |
+    #       (+authors.full_name:schwartz +title:schwartz))"
+    extra_params["type"] = "cross_fields"
 
 
 ###############################################################################
--- a/tests/api/test_documents_search.py
+++ b/tests/api/test_documents_search.py
@@ -1,0 +1,88 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2024 CERN.
+#
+# CDS-ILS is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Tests search OR/AND operators.
+
+These tests asserts that when using the AND operator, the `cross_fields`
+parameter is required. The `Dreams` word is contained in the `title`, the
+`Caroline` word in the author.
+The assumption here is that the documents index has the same analyzers
+for all searchable fields.
+"""
+
+from copy import deepcopy
+
+import pytest
+from flask import url_for
+
+from cds_ils import config
+
+
+@pytest.fixture(scope="module")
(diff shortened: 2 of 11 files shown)


sakshamarora1 (commented):
The original example mentioned in the inline comment wasn't the complete case. I updated with the correct explanation according to my testing and observations.

Essentially, we can not use cross_fields with AND operator as the operator forces both the terms to be in the same field which means it ignores the cross_fields type.

If we remove the AND operator, it works but will increase the number of results that would loosely match due to the default OR.

The best way I was able to find was to inject "AND" keyword between each word, this seems to solve the problem completely but it is not an elegant solution.

zzacharo:
Here some tests in devtools of opensearch (from @ntarocco ):

```
# 0s8wv-bxz77
GET /documents/_search
{
  "profile": true,
  "query": {
    "query_string": {
      "query": "quantum AND schwartz",
      "fields": ["title", "authors.full_name"]
    }
  }
}

GET /documents/_search
{
  "profile": false,
  "query": {
    "query_string": {
      "query": "quantum schwartz",
      "type": "cross_fields",
      "default_operator": "AND"
    }
  }
}

GET /documents/_search
{
  "profile": true,
  "query": {
    "query_string": {
      "query": "quantum field theory schwartz",
      "type": "cross_fields",
      "default_operator": "AND",
      "fields": ["title", "authors.full_name"]
    }
  }
}

GET /documents/_search
{
  "profile": false,
  "query": {
    "bool": {
      "must": [
        {"query_string": {
      "query": "field schwartz",
      "type": "cross_fields",
      "default_operator": "AND",
      "fields": ["title", "*"]
    }},
    {"term": {
      "pid": {
        "value": "0s8wv-bxz77"
      }
    }}
    
      ]
    }
    
  }
}
```

zzacharo:
Working query
```
GET /documents/_search 
{
  "profile": true,
  "query": {
    "query_string": {
      "query": "quantum schwartz",
      "type": "cross_fields",
      "default_operator": "AND",
    "analyzer": "standard"
    }
  }
}
```

# Findings

Cross field queries: https://www.elastic.co/guide/en/elasticsearch/guide/current/_cross_fields_queries.html

> For the cross_fields query type to work optimally, all fields should have the same analyzer. Fields that share an analyzer are grouped together as blended fields.
> If you include fields with a different analysis chain, they will be added to the query in the same way as for best_fields. For instance, if we added the title field to the preceding query (assuming it uses a different analyzer), the explanation would be as follows:
> (+title:peter +title:smith)

It seems that fields `title` and `authors.full_name` use a different analyzer. Forcing the analyzer on the query level as above seems to make the `cross_fields` type work as expected.

tests/api/test_documents_search.py
```diff
@@ -0,0 +1,40 @@
...
+
+    assert response.status_code == 200
+    result = response.get_json()
+    assert result["hits"]["total"] == 1
+
+
+def _query_params_modifier(extra_params={}):
+    extra_params["default_operator"] = "AND"
+    extra_params["type"] = "cross_fields"
+    # FIXME: this seems not to be taken into account
```
kpsherva, line 26: is this comment for the line above?

cds_ils/config.py
```diff
@@ -87,8 +87,23 @@ def _parse_env_bool(var_name, default=None):
...
+    # "query_string": {
+    #   "query": "Theory Schwartz",
+    #   "type": "cross_fields",
+    #   "fields": ["title", "authors.full_name"]
+    #   "analyzer": "custom_analyzer"
+    # }
+    # Explanation:
+    # "((+title:theory +authors.full_name:theory) | (+authors.full_name:schwartz +title:schwartz))"
+    extra_params["type"] = "cross_fields"
+    extra_params["analyzer"] = "custom_analyzer"
```
kpsherva, line 105: I am thinking... aren't we losing the possibility to search in all of the other fields which don't have the custom_analyzer? so all the fields other than author and title?, Ping @sakshamarora1

sakshamarora1: Yes, this was also what I was worried about.
Just checked and confirmed: `document_type:MULTIMEDIA` will become `document_type:multimedia` which won't return any results.

Issue #923, opened 2024-11-01
serials importer: broken matching by titles


anikachurilova:
When importing serials, title normalizer is supposed to remove substrings `[" series", " ser", " Series", " Ser"]` from **the end of the title**.
Currently, it is removing those substrings from the whole string, resulting in such normalised titles:
- `"Springer Series in Statistics Series"` --> `"springeries in statistics"` (expected value is `"springer series in statistics"`)
- `"International   series of Numerical mathematics   ser"` --> `"internationalies of numerical mathematics"` (expected value is `"international series of numerical mathematics"`)

As a result - no matches found by the broken title in the existing serials and the serial with the same title gets imported as a new one.

PR #925, opened 2024-11-01
serials importer: remove substrings only from the end of the title

anikachurilova:
* closes https://github.com/CERNDocumentServer/cds-ils/issues/923

### DIFF
--- a/cds_ils/importer/series/importer.py
+++ b/cds_ils/importer/series/importer.py
@@ -120,7 +120,8 @@ class SeriesImporter(object):
         # will become
         # `international series of numerical mathematics`
         for substring in IGNORE_SUFFIXES:
-            t = rreplace(t, substring, "")
+            if re.search(substring + "$", t):
+                t = rreplace(t, substring, "")
         return t.strip()
 
     def update_series(self, matched_series, json_series):


cds_ils/importer/series/importer.py
```diff
@@ -120,7 +120,7 @@ def _normalize_title(title):
         # will become
         # `international series of numerical mathematics`
         for substring in IGNORE_SUFFIXES:
-            t = rreplace(t, substring, "")
+            t = rreplace_end(t, substring)
```
kpsherva, line 123: I think it would be better to enhande rreplace function to use regex, as in
```
regex = substring + "$" # if i remember correctly this is the mark of the end of the string
if re.search(regex, series["title"]):
 series["title"] = rreplace(series["title"], regex, "")
```
instead of introducing a new parser function
please check also this occurrence:

https://github.com/CERNDocumentServer/cds-ils/commit/a7fd72e237e8ead29ad3a105d59d6949a192cc9a#diff-49df47ccd415564e19fca026f91495c2919495215eabea18c8bd00f31dd65049R88

sakshamarora1: I think putting `Series` before `ser` in the list might solve this no? @anikachurilova

PR #936, opened 2025-03-19
importer: add matching rules for DOIs, standard numbers and URLs for videos

anikachurilova:


cds_ils/importer/documents/importer.py
```diff
@@ -322,7 +365,18 @@ def validate_found_matches(self, not_validated_matches):
...
+            # match videos by URL
+            import_eitem = self.json_data.get("_eitem", {})
+            if import_eitem.get("_type") == "video":
+                eitem_search = current_app_ils.eitem_search_cls()
+                eitem_cls = current_app_ils.eitem_record_cls
+                document_eitems = eitem_search.search_by_document_pid(pid_value)
+                for hit in document_eitems:
+                    document_eitem = eitem_cls.get_record_by_pid(hit.pid)
+                    if self._matching_video_urls(import_eitem, document_eitem):
+                        matches.append(pid_value)
```
kpsherva, line 377: from what I undestood Lydia was indicating in the ticket, they would like to search by video url to match document. You are validating already found matches if they have correct value. This is good, but we are missing the search part

kpsherva (approved):

PR #995, opened 2025-06-02
install: added documentation for how to install+start ui

JakobMiesner:


INSTALL.rst
```diff
@@ -65,13 +71,20 @@ and after the setup script:
 
 Running
 -------
-Start the webserver and the celery worker:
+Start the backend and the celery worker:
```
kpsherva, line 74: nothing to do with the PR but the text says we start celery too, which is not the case, can you add the celery command below?
`./scripts/celery`

JakobMiesner: done

kpsherva (approved):

Issue #879, opened 2024-04-05
add information about other e-items of the same provider


kpsherva:
On the import time (or delete) e-items are matched by the creator value (in this case provider, value is inserted on import time), see
https://github.com/CERNDocumentServer/cds-ils/blob/master/cds_ils/importer/eitems/api.py#L15
this workflow won't work if the e-item was created manually by the librarian - leading to duplications and not deleted e-items. We want to avoid using "source" field of the e-item to match them due to possible human errors while editing this field.
Instead we should add an information in the report stating that another e-item of the currently processed provider was detected and manual action is required.
The information should be visible in "Actions and details" as well as the row should be highlighted in yellow

![Image](https://github.com/CERNDocumentServer/cds-ils/assets/38131488/21cbac1a-aa04-4b07-ba1d-8616718c6650)

lydipi:
@kpsherva I just did, and it only update document x0tap-26g10 without creating a new one 👍 so it seems only the label is wrong.

However, one thing I noticed is that it created a new e-item because the source is _different_ but it should be the same: when manually creating an item and selecting the source, it is "SAFARI" (= O'Reilly): https://sandbox-cds-ils.web.cern.ch/backoffice/eitems/6dfwx-h4e85. When coming from a Safari import, it is "safari": https://sandbox-cds-ils.web.cern.ch/backoffice/eitems/frpzb-xrr41. As both come from source "safari", it should be the same source, independently if it was created manually or from import.

lydipi:
@kpsherva just to be sure I understood correctly, then if we would like to avoid having a second e-item created, we would have to delete the record from the XML before doing the import?

PR #1012, opened 2025-07-02
importer: add information about other e-items of the same provider

JakobMiesner:
closes: https://github.com/CERNDocumentServer/cds-ils/issues/879


cds_ils/importer/serializers/schema.py
```diff
@@ -22,6 +22,7 @@ class ImportedEItemSchema(Schema):
     action = fields.String(dump_only=True)
     priority_provider = fields.Bool()
     duplicates = fields.List(fields.String)
+    ambiguous = fields.List(fields.String)
```
kpsherva, line 25: I am a bit unsure what would be an ambiguous e-item in the context, let's chat about the details

PR #1095, opened 2026-06-17
fix(document details): browser history modification

kpsherva:
* fixes https://cern.service-now.com/nav_to.do?uri=incident.do?sysparm_query=number=INC4876597

### DIFF
--- a/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculationExtras.js
+++ b/ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculationExtras.js
@@ -1,10 +1,38 @@
-import React from "react";
+import React, { useRef, useEffect } from "react";
 import _get from "lodash/get";
 import _isEmpty from "lodash/isEmpty";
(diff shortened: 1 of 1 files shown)


ui/src/overridden/frontsite/Document/DocumentDetails/DocumentCirculationExtras.js
```diff
@@ -1,11 +1,39 @@
...
+// eslint-disable-next-line react/prop-types
+function MapIframe({ url }) {
+  const iframeRef = useRef(null);
+  useEffect(() => {
+    if (iframeRef.current) {
+      iframeRef.current.contentWindow.location.replace(url);
+    }
+  }, [url]);
+  return (
+    <div className="ui active embed cern-map">
```
sakshamarora1, line 20: Was the`<Embed>` component not working with replace url?

kpsherva: the implementation of embed was causing the redundant entries added to the history of the browser, like the comment over the new function mentions

sakshamarora1 (approved):

PR #1099, opened 2026-07-15
change(setup): update dependencies

callumlewis05:


ui/package.json
```diff
@@ -45,6 +45,9 @@
     "@semantic-ui-react/craco-less": "^3.0.0",
     "react-scripts": "^5.0.1"
   },
+  "overrides": {
```
kpsherva, line 48: was there any error that this change was needed for?

PR #1102, opened 2026-08-03
fix(importer): stopped incorrect persisting error messages

callumlewis05:
If preview was clicked before all fields had been filled the error messages would display and then persist even when the field had been filled but now disappear once the input is clicked on.

* closes https://github.com/inveniosoftware/react-invenio-app-ils/issues/700

### DIFF
--- a/ui/src/importer/Importer.js
+++ b/ui/src/importer/Importer.js
@@ -39,9 +39,16 @@ export class Importer extends Component {
   }
 
   handleChange = (e, { name, value }) => {
-    this.setState({
-      [name]: value,
-    });
+    const update = { [name]: value };
+
+    if (name === "provider") {
+      update.providerMissing = false;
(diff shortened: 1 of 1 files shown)


kpsherva (approved):
if we find ourselves adding more make-shift validation here, maybe it makes sense to add a library which handles a form out of the box, including validation.
I don't remember, do we have a rest API for this feature? is the backend validating the input? I assume you only fixed the frontend because of the state kept in the components, correct?

callumlewis05:
This change only edits how it visually looks, I do not think extra validation is needed as it is only from a drop down menu but the validation on clicking preview is ensuring that required fields are filled, which would then not disappear once the fields were being edited, though now returns to normal when the dropdown is opened.

kpsherva (approved):

